PR #2, opened 2018-09-25
UIDATIMP-2: Adjust skeleton application.

VictorSoroka1:


package.json
```diff
@@ -12,37 +15,45 @@
     "start": "stripes serve",
     "build": "stripes build --output ./output",
     "test": "stripes test nightmare --run demo --show",
-    "lint": "eslint src"
+    "lint": "eslint ."
```
oleksii-popov: Please add --fix

VictorSoroka1: I would prefer to not do this. As `lint` task is firing on CI which means some errors could not be spotted during that check. I could add just additional script like `lintfix` for that purpose.

PR #3, opened 2018-10-01
Moddataimp 6: Implement search panel

OleksandrAntonenko1:


src/components/search-panel/search-panel.css
```diff
@@ -0,0 +1,47 @@
+.nested-container {
+    margin-left: 1rem
+}
+
+.container {
+    margin-left: 1.5rem
+}
+
+.record-type {
+    height: 10px;
```
VictorSoroka1: Let's follow the attribute ordering.
1. float, positioning (if needed)
2. display (if flex add below flex rules)
3. layout (width, height, margin, padding, border - in that order)
3. font rules
4. others

src/components/search-panel/search-panel.js
```diff
@@ -0,0 +1,219 @@
...
+        id="ex-1"
+        separator={false}
+        header={FilterAccordionHeader}
+      >
+        <div className={css['nested-container']}>
+          <Checkbox
+            name="checkbox_1"
+            label="Succesful (25)"
+          />
+          <div className={css['nested-container']}>
```
VictorSoroka1: Name classes in camelcase and then refer to them in the following way - `css.nestedContainer`. This adds some sort of lint checking.

PR #6, opened 2018-10-05
MODDATAIMP-12: Fix result block behavior.

OleksandrAntonenko1:


src/settings/index.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import Settings from '@folio/stripes-components/lib/Settings';
+import Settings from '@folio/stripes-smart-components/lib/Settings';
```
VictorSoroka1: Please change the import to be `{ Settigns } from ...`.

src/components/ResultPanel/ResultPanel.css
```diff
@@ -39,6 +40,8 @@
 
 .accordionHeader {
   width: 100%;
+  flex-wrap: wrap;
```
VictorSoroka1: Let's have it above the width.

src/components/ResultPanel/ResultPanel.css
```diff
@@ -2,6 +2,7 @@
 
 .recordType {
   display: inline-block;
+  flex-shrink: 0;
```
VictorSoroka1: Let's have above the display. BTW, do we need display there as it seems to be flex item?

src/components/ResultPanel/ResultPanel.css
OleksandrAntonenko1: Done.

src/components/ResultPanel/ResultPanel.css
OleksandrAntonenko1: Done.

PR #12, opened 2018-10-18
UIDATIMP-22: Add "Jobs" pane layout.

VictorSoroka1:


.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "singleQuote": true,
```
oleksii-popov: Do we actually use Pretier?

src/index.js
```diff
@@ -13,14 +13,24 @@ class DataImport extends React.Component {
     showSettings: PropTypes.bool,
   };
 
+  constructor(props) {
+    super(props);
+
```
oleksii-popov: please remove empty lines

src/index.js
```diff
@@ -13,14 +13,24 @@ class DataImport extends React.Component {
...
       return <Settings {...this.props} />;
     }
 
     return (
       <Switch>
-        <Route path={`${this.props.match.path}`} exact component={Application} />
+        <Route
+          path={`${this.props.match.path}`}
+          exact
+          render={() => <this.connectedHome {...this.props} />}
```
oleksii-popov: PLease move to additional method and add comments.

src/index.js
VictorSoroka1: I usually put an empty line after the `super`. So this one is not accidental.

src/index.js
VictorSoroka1: Will do.

.prettierrc
VictorSoroka1: I personally use it for sure while development. It is up to the developer to enable it for the redactor/IDE.

PR #13, opened 2018-10-25
UIDATIMP-31: Add reusable Job component.

VictorSoroka1:


src/components/Jobs/Jobs.js
```diff
@@ -1,26 +1,51 @@
...
-        />
-      </AccordionSet>
+      <div className={css.jobsPane}>
+        <AccordionSet>
+          <Accordion
+            label={formatMessage({ id: 'ui-data-import.previewJobs' })}
+            separator={false}
+          >
+            <div className={css.jobList}>
+              <Job job={jobsMocks[0]} checkDateIsToday={this.checkDateIsToday} />
```
maximdidenkoepam: Please move each property to a new line. The same issue is in line 43

VictorSoroka1: As I understand you consider best practice to move properties on a new line if it more that one property, right?

src/components/Jobs/components/Job/Job.js
```diff
@@ -0,0 +1,112 @@
...
+
+import Progress from '../Progress';
+import jobMetaTypes from './jobMetaTypes';
+
+import css from './Job.css';
+
+class Job extends React.Component {
+  static defaultProps = {
+    handlePreview: () => {
+      // TODO: to be implemented in further stories
```
maximdidenkoepam: Can we use _.noop for this case?

VictorSoroka1: Nice idea. Will do.

src/components/Jobs/components/Job/Job.js
```diff
@@ -0,0 +1,112 @@
...
+    const {
+      jobProfileName,
+      fileName,
+      status,
+      jobExecutionHRID,
+      runBy: { firstName, lastName },
+      progress: { current, total },
+    } = job;
+    const jobMeta = jobMetaTypes[status](job);
+    const dateLabel = `ui-data-import.${jobMeta.dateLabel}Running`;
```
maximdidenkoepam: As I understand it is not a label. It is a key or translation id. Maybe we should rename it?

VictorSoroka1: Do you mean to remove variable and put it like that? If so, I agree.

maximdidenkoepam: I am ok with variable, but it is not a label, it is label id or key. I suggest to rename it for example to dateLabelId?

VictorSoroka1: Agree, will rename.

src/components/Jobs/components/Job/Job.css
```diff
@@ -0,0 +1,32 @@
...
+  overflow-wrap: break-word;
+
+  & > span:not(:last-child):after {
+    content: " \B7 ";
+  }
+}
+
+.job {
+  padding: .5rem;
+  border: 1px solid var(--color-border);
```
oleksii-popov: Good usage of variables

src/components/Jobs/Jobs.js
```diff
@@ -1,26 +1,86 @@
...
 import { stripesShape } from '@folio/stripes/core';
-import { AccordionSet, Accordion } from '@folio/stripes/components';
+import {
+  AccordionSet,
+  Accordion,
+} from '@folio/stripes/components';
+
+import Job from './components/Job';
+import EndOfList from './components/EndOfList';
+import jobsMocks from './jobsMocks'; // TODO: to be replaced with real data in further stories
```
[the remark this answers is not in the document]
VictorSoroka1: Nice idea. Will do.

PR #14, opened 2018-10-26
UIDATIMP-33: Add UI for file upload.

VictorSoroka1:


src/components/ImportJobs/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,35 @@
...
+      title,
+      uploadBtnText,
+    } = this.props;
+
+    return (
+      <div className={css.upload}>
+        <span className={css.uploadTitle}>{title}</span>
+        <Button
+          buttonStyle="primary"
+          onClick={noop}
```
oleksii-popov: Do we really need to include _ noop? can we simple use javascript void?

VictorSoroka1: I think it is good to be consistent with such things. As this way is already used in another part of the application.

PR #17, opened 2018-11-09
UIDATIMP-29: Add UI for jobs logs pane.

VictorSoroka1:


src/components/JobLogs/withJobLogsCellsFormatter.js
```diff
@@ -0,0 +1,59 @@
...
+  intlShape,
+} from 'react-intl';
+
+import { compose } from '../../utils';
+
+const withJobLogsCellsFormatter = WrappedComponent => {
+  return class extends Component {
+    static propTypes = {
+      intl: intlShape.isRequired,
+      formatter: PropTypes.object,
```
maximdidenkoepam: Do we need default value for this property?

src/components/JobLogs/withJobLogsCellsFormatter.js
```diff
@@ -0,0 +1,59 @@
...
+
+      this.cellFormatters = {
+        ...props.formatter,
+        runBy: this.formatUser,
+        completedDate: this.formatEndedRunningDate,
+      };
+    }
+
+    formatUser({
+      runBy: {
```
maximdidenkoepam: Please use class property approach. Also in this case I suggest to move destructuring from declaration to function body.

src/components/JobLogs/withJobLogsSort.js
```diff
@@ -0,0 +1,164 @@
...
+  sortNums,
+  sortDates,
+  sortStrings,
+} from '../../utils/sort';
+
+const withJobLogsSort = WrappedComponent => {
+  return class extends Component {
+    static propTypes = {
+      intl: intlShape.isRequired,
+      formatter: PropTypes.object,
```
maximdidenkoepam: Please add missed props to default props

src/components/JobLogs/withJobLogsSort.js
```diff
@@ -0,0 +1,164 @@
...
+      if (isSameField) {
+        sortDir = direction === sortDir ? 'descending' : sortDir;
+      }
+
+      const sortState = {
+        sort: fieldName,
+        direction: sortDir,
+      };
+
+      this.setState(() => sortState);
```
maximdidenkoepam: `this.setState(sortState)`?

src/routes/Home.js
```diff
@@ -1,16 +1,39 @@
...
 
 import Jobs from '../components/Jobs';
+import JobLogs from '../components/JobLogs';
 import ImportJobs from '../components/ImportJobs';
 
-export default class Home extends React.Component {
+class Home extends Component {
   static propTypes = {
     stripes: stripesShape.isRequired,
+    resources: PropTypes.shape({
```
maximdidenkoepam: Please add default props.

src/components/JobLogs/withJobLogsSort.js
```diff
@@ -0,0 +1,164 @@
...
+        direction = 'ascending',
+      } = qs.parse(this.props.location.search.slice(1));
+
+      this.setState({
+        sort,
+        direction,
+      });
+    }
+
+    prepareLogsData(logs = []) {
```
maximdidenkoepam: I suggest use class props as much as possible. Please update this component.

src/components/JobLogs/withJobLogsSort.js
VictorSoroka1: I did this way intentionally. I prefer this way of setting state.

https://medium.com/@wisecobbler/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1

src/components/JobLogs/withJobLogsCellsFormatter.js
VictorSoroka1: Do you mean to use static?

PR #18, opened 2018-11-12
Replace formatMessage() with FormattedMessage

cherewaty:
Related to https://github.com/folio-org/ui-circulation/pull/124

`stripes.intl` is deprecated. Preferred approach:
Use `<FormattedMessage>`, `<FormattedDate>`, and `<FormattedTime>` as often as possible, only falling back to `formatMessage()`, `formatDate()`, and `formatTime()` (accompanied by `injectIntl`) when a bare string is absolutely necessary.

I left one instance of `formatDate()` where a date comparison was happening, but that function should be re-evaluated - the date comparison can probably take place without localizing the date strings.

### DIFF
--- a/src/components/Jobs/Jobs.js
+++ b/src/components/Jobs/Jobs.js
@@ -1,6 +1,6 @@
 import React from 'react';
+import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
 
-import { stripesShape } from '@folio/stripes/core';
 import {
   AccordionSet,
   Accordion,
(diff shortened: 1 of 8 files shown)


src/components/Jobs/Jobs.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
+import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
```
OleksandrAntonenko1: In Epam we are using such code style. Feel free to join to our side)
```suggestion
import { 
  FormattedMessage, 
  injectIntl, 
  intlShape,
} from 'react-intl';
```

cherewaty: Do y'all have a strict rule about when to make the import multi-line vs. single-line? Is it just any list of imports greater than one is multi-line? (I'll happily follow your conventions on that, I haven't been consistent about it.)

OleksandrAntonenko1: Yeah you right. It Is just any list of imports greater than one is multi-line.

PR #20, opened 2018-11-12
UIDATIMP-29: Use formattedMessage instead of FormattedMessage compone…

VictorSoroka1:
…nt because of MultiColumnList component peculiarities with calculations for column widths.

### DIFF
--- a/src/components/JobLogs/JobLogs.js
+++ b/src/components/JobLogs/JobLogs.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { FormattedMessage } from 'react-intl';
+import { intlShape } from 'react-intl';
 
 import {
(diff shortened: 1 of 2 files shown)


src/components/JobLogs/JobLogs.js
```diff
@@ -89,4 +92,7 @@ class JobLogs extends Component {
   }
 }
 
-export default compose(withJobLogsCellsFormatter, withJobLogsSort)(JobLogs);
+export default compose(
+  withJobLogsCellsFormatter,
+  withJobLogsSort,
+)(JobLogs);
```
cherewaty: Oh, do you need to `injectIntl` here?

VictorSoroka1: No, it is provided in `withJobLogsCellsFormatter`. So there is no need to inject it twice.

PR #25, opened 2018-11-16
UIDATIMP-52

OleksandrYehorovEPAM:
Changed contracts according to /metadata-provider/jobExecutions endpoint.


src/components/JobLogs/JobLogs.js
```diff
@@ -86,6 +86,7 @@ class JobLogs extends Component {
         formatter={formatter}
         sortOrder={this.columnMapping[sortField]}
         sortDirection={sortDirection}
+        autosize
```
VictorSoroka1: Why do I see three commits in the history whereas I only expect one? Please check this.

PR #27, opened 2018-11-22
UIDATIMP-37: Create UI for File Upload: Action & Styling for "In Process" file uploads

Nick-Yenin:
Has been added a new route and UploadingDisplay component. Implemented feature of uploading files to the server.


src/components/UplaodingDisplay/components/FileItem/FileItem.js
```diff
@@ -0,0 +1,42 @@
...
+      name,
+      currentUploaded,
+      size,
+    } = this.props;
+
+    return (
+      <div className={css.fileItem}>
+        <span>{name}</span>
+        <Progress
+          message="Uploading"
```
anton-plotnikov: Please, check do we use this `message` as a message displayed to the user and if so - use `<FormattedMessage>` with translation

src/components/UplaodingDisplay/UploadingDisplay.js
```diff
@@ -0,0 +1,254 @@
...
+    endpointFileUpload: PropTypes.string,
+  };
+
+  static defaultProps = {
+    endpointFileDef: '/data-import/upload/definition',
+    endpointFileUpload: '/data-import/upload/file',
+  };
+
+  static mapFilesFromProps(props, reducer, initReducerValue = {}) {
+    if (!props.initState) return {};
```
VictorSoroka1: Add `defaultProp` for that so this line can be removed.

src/components/UplaodingDisplay/UploadingDisplay.js
```diff
@@ -0,0 +1,254 @@
...
+
+    return props.initState.reduce(reducer, initReducerValue);
+  }
+
+  static mapFilesFromPropsReducer(result, currentFile) {
```
VictorSoroka1: Let's combine `mapFilesFromProps` and `mapFilesFromPropsReducer` in one function.

src/components/UplaodingDisplay/UploadingDisplay.js
```diff
@@ -0,0 +1,254 @@
...
+    const keyNameValue = currentFile.name + currentFile.lastModified;
+
+    currentFile.keyName = keyNameValue;
+    currentFile.currentUploaded = 0;
+    result[keyNameValue] = currentFile;
+
+    return result;
+  }
+
+  static addHeaders = (xhr, headersObj) => {
```
VictorSoroka1: Rename to be just `headers`. We should avoid naming arguments with type in it as it can be changed later.

src/components/UplaodingDisplay/UploadingDisplay.js
```diff
@@ -0,0 +1,254 @@
...
+
+  static async getUploadDefinition(fullUrl, config) {
+    return fetch(fullUrl, config)
+      .then(res => res.json());
+  }
+
+  state = UploadingDisplay.mapFilesFromProps(this.props, UploadingDisplay.mapFilesFromPropsReducer);
+
+  async componentDidMount() {
+    const { endpointFileDef, endpointFileUpload } = this.props;
```
VictorSoroka1: Follow that pattern for all similar cases per our convention
```
    const {
      history,
      match,
    } = this.props;
```

src/components/Progress/Progress.js
```diff
@@ -0,0 +1,67 @@
...
+    message,
+    progressClassName,
+    progressWrapperClassName,
+    progressInfoClassName,
+    progressCurrentClassName,
+  } = props;
+  const progressValue = calculatePercentage(current, total);
+  let progressInfo;
+
+  if (progressInfoType === 'messagedPercentage') {
```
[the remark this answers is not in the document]
Nick-Yenin: I think it is not necessary in this particular case.

src/components/UplaodingDisplay/UploadingDisplay.js
```diff
@@ -0,0 +1,254 @@
...
+    const headerKeys = Object.keys(headersObj);
+
+    headerKeys.forEach(headerKey => {
+      xhr.setRequestHeader(headerKey, headersObj[headerKey]);
+    });
+
+    return xhr;
+  };
+
+  static prepareFilesDefinition(filesToUploadObj) {
```
VictorSoroka1: Let's refactor the body with the following usage
```
    return Object.keys(filesToUpload)
      .reduce((files, currentKey) => files.concat({ name: currentKey }), [])
      .map(files => ({ fileDefinitions: files }));
```

PR #29, opened 2018-11-28
UIDATIMP-5: Add Data Import settings page.

VictorSoroka1:


src/settings/DataImportSettings.js
```diff
@@ -0,0 +1,93 @@
+import React, { Component } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import { Settings } from '@folio/stripes/smart-components';
+import { AppIcon, InfoPopover } from '@folio/stripes/components';
```
OleksandrYehorovEPAM: Please, put each item on its own line

src/settings/ActionProfiles.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+
+import { Pane } from '@folio/stripes/components';
+
+export default class ActionProfiles extends Component {
```
OleksandrYehorovEPAM: Would it be better to just create another component or HOC to get rid of code reuse in such components as ActionProfiles, FieldMappingProfiles, FileExtensions and others? Seems like all of those components are having the same code except FormattedMessage. Please correct me if I am wrong.
And also maybe it would be better to use SFC (functional component) as there is no hooks and state in these components.

src/settings/DataImportSettings.js
```diff
@@ -0,0 +1,93 @@
...
+import ActionProfiles from './ActionProfiles';
+import FieldMappingProfiles from './FieldMappingProfiles';
+import FileExtensions from './FileExtensions';
+
+import css from './DataImportSettings.css';
+
+class DataImportSettings extends Component {
+  sections = [
+    {
+      label: this.renderProfilesLabel(),
```
oleksii-popov: i suggest to have routes as separated 'routes' files

src/settings/DataImportSettings.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+  ];
+
+  renderProfilesLabel() {
+    return (
+      <div className={css.profilesLabel}>
+        <FormattedMessage id="ui-data-import.settings.profiles" />
+        <InfoPopover
+          allowAnchorClick
+          buttonHref="https://wiki.folio.org/display/FOLIOtips/Creating+and+using+profiles"
```
oleksii-popov: Please move out of here please

src/settings/DataImportSettings.js
```diff
@@ -0,0 +1,93 @@
...
+        app="data-import"
+        iconKey={iconKey}
+      >
+        <FormattedMessage id={`ui-data-import.settings.${labelId}`} />
+      </AppIcon>
+    );
+  }
+
+  render() {
+    return (
```
oleksii-popov: WHy do we need a return in render function? As I see component is written in Class notation.

src/settings/DataImportSettings.js
VictorSoroka1: Will do.

src/settings/ActionProfiles.js
VictorSoroka1: Discussed it. Leave as is for now.

src/settings/DataImportSettings.js
VictorSoroka1: I think this is `Settings` component is just for it.

src/settings/DataImportSettings.js
VictorSoroka1: Please see the documentation for this.
`When called, it should examine this.props and this.state and return one of the following types...`

https://reactjs.org/docs/react-component.html#render

src/settings/DataImportSettings.js
VictorSoroka1: Personally, I do not see a reason for doing this as this link address is bound to profiles label and in the case when it should be changed this the exact place to edit it.

PR #35, opened 2018-12-09
UIDATIMP-63: Add UI for file uploading fail.

VictorSoroka1:
## Purpose
To have UI for error handling, when something goes wrong while files are uploading (e.g. timeout, communication problem)

## Approach
Extend `FileItem` component to support multiple states like `uploading`, `uploaded` and `failed` and differentiate behavior based on that. Additionally, cover both error cases: either error during posting upload definition or error on individual file error.

## Screenshots
screen shot 2018-12-10 at 12 18 14 am

### DIFF
--- a/src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
+++ b/src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
@@ -1,20 +1,94 @@
-import React, { PureComponent } from 'react';
+import React, {
+  PureComponent,
+  Fragment,
+} from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
 
+import {
+  Icon,
+  IconButton,
+} from '@folio/stripes/components';
+
+import {
+  UPLOADED,
+  UPLOADING,
+  FAILED,
+} from './fileItemStatuses';
 import Progress from '../../../Progress';
 
 import css from './FileItem.css';
+
+const getFileItemMeta = (props) => {
+  const {
+    uploadStatus,
+    name,
+  } = props;
+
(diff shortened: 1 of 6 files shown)


translations/ui-data-import/en.json
```diff
@@ -20,13 +20,15 @@
...
   "undo": "Undo",
+  "delete": "Delete",
 
   "uploadTitle": "Drag and drop to start a new data import job",
   "activeUploadTitle": "Drop to start import",
   "uploadBtnText": "or choose files",
   "uploadingPaneTitle": "Files",
   "uploadingMessage": "Uploading",
-  
+  "uploadFileError": "Error: Upload file",
```
oleksii-popov: Error uploading file

VictorSoroka1: Going to raise the question to Ann-Marie about this. The current message is retrieved from mockup and for the moment is the source of truth.

src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
```diff
@@ -23,23 +97,34 @@ class FileItem extends PureComponent {
...
-          total={size}
-          current={uploadedValue}
-        />
+      <div className={classNames(css.fileItem, meta.fileWrapperClassName)}>
+        <div className={css.fileItemHeader}>
+          {meta.renderHeading()}
+        </div>
+
+        {meta.showProgress &&
+          <Progress
```
OleksandrYehorovEPAM: Consider wrapping it with parenthesis as it is a multiline JSX expression.

VictorSoroka1: Agree as long as we act consistent with this. Should be okay now.

PR #41, opened 2018-12-13
UIDATIMP-38: Create UI for styling for `Completed` file uploads

VictorSoroka1:
## Purpose

Create UI for styling for Completed file uploads

## Approach

Extend `FileItem` component to support uploaded and deleting file uploading states. Additionally, cover error case when file removal was not succeded. In this case, the state should be changed back to the uploaded state.


src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -1,62 +1,88 @@
...
-  createFileDefinition, // eslint-disable-line
-  prepareFilesToUpload, // eslint-disable-line
-  uploadFiles,
-} from './utils/upload';
+import createUrl from '../../utils/createUrl';
+import * as API from './utils/upload';
+
+const DEFAULT_TIMEOUT_BEFORE_FILE_DELETION = 0;
+
+const NoUploadedFilesMessage = () => (
```
maximdidenkoepam: This function is used once, can we move the implementation to the place of the invocation?

src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
```diff
@@ -3,27 +3,30 @@ import React, {
...
-  UPLOADING,
-  FAILED,
-} from './fileItemStatuses';
+import * as fileStatuses from './fileItemStatuses';
 import Progress from '../../../Progress';
 
 import css from './FileItem.css';
 
-const getFileItemMeta = (props) => {
+const getFileItemMeta = props => {
```
maximdidenkoepam: I suggest to move getFileItemMeta from the file to another place, and then import it and use in the component

src/components/UploadingJobsDisplay/utils/upload.js
```diff
@@ -119,3 +119,29 @@ export const uploadFiles = (
...
+    const error = {
+      status,
+      statusText,
+    };
+
+    throw new Error(error);
+  }
+};
+
+export const deleteFile = (url, headers) => {
```
maximdidenkoepam: Can we use async/await here?

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -127,19 +165,58 @@ class UploadingJobsDisplay extends Component {
...
+          console.error(error); // eslint-disable-line no-console
+        });
+    }, timeoutBeforeFileDeletion);
+
+    this.updateFileState(file, { status: fileStatuses.DELETING });
+  };
+
+  deleteFileFromState = key => {
+    const { files } = this.state;
+    const updatedFiles = Object.assign({}, files);
```
OleksandrYehorovEPAM: Consider rewriting with spread operator instead
```suggestion
    const updatedFiles = { ...files };
```

src/components/UploadingJobsDisplay/utils/upload.js
VictorSoroka1: I would keep using as is as this approach is used in other places in this file. Moreover, we have not decided on strictly using async/await syntax.

src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
VictorSoroka1: Makes sense. Will do.

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
VictorSoroka1: I thought about it. I think it makes sense eventually, so will do this.

src/components/Progress/Progress.js
```diff
@@ -10,7 +10,7 @@ const progressInfoFormatters = {
   messagedPercentage: (current, total, payload) => (
     <React.Fragment>
```
OleksandrYehorovEPAM: Consider using named import for Fragment
```suggestion
    <Fragment>
          // ...
    </Fragment>
```

src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
```diff
@@ -3,27 +3,30 @@ import React, {
   Fragment,
 } from 'react';
 import PropTypes from 'prop-types';
-import { FormattedMessage } from 'react-intl';
+import {
+  FormattedMessage,
+  FormattedDate,
+} from 'react-intl';
 import classNames from 'classnames';
+import noop from 'lodash/noop';
```
OleksandrYehorovEPAM: Let's be consistent and use named import.
```suggestion
import { noop } from 'lodash';
```

src/components/Progress/Progress.js
VictorSoroka1: Agree with that.

src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
VictorSoroka1: Will do.

PR #42, opened 2018-12-17
UIDATIMP-67: Prevent user navigation when file upload is in progress

oleksii-popov:
**Purpose**
When the upload is in progress - prevent user navigation out of the page (or reload a page). In this case upload process will not be interrupted.

**Ticket**
[UIDATIMP-67](https://issues.folio.org/browse/UIDATIMP-67)

**Screenshot**
leave site


src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -48,6 +48,29 @@ class UploadingJobsDisplay extends Component {
     this.uploadJobs();
   }
 
+  componentDidUpdate() {
+    if (!this.props.files) {
```
VictorSoroka1: Let's move all this stuff to the method as `componentDidUpdate` is just a hook which can operate on different cases.

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -48,6 +48,29 @@ class UploadingJobsDisplay extends Component {
...
+      return;
+    }
+
+    let dontLeave = false;
+
+    for (const file in this.props.files) {
```
VictorSoroka1: Let's use the code below instead of `for in` loop.
```suggestion
    const isAbleToLeavePage = Object.keys(this.props.files)
      .some(file => file.status === fileStatuses.UPLOADING);
```

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -48,6 +48,29 @@ class UploadingJobsDisplay extends Component {
...
+      if (this.props.files[file].status.includes('Uploading')) {
+        dontLeave = true;
+        break;
+      }
+    }
+    // prevent from leaving the page in case of download in progress
+    if (dontLeave) {
+      window.onbeforeunload = () => {
```
VictorSoroka1: This code does not cover the case when there is the file in process of uploading and the user navigates away to another page within the folio application.

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -48,6 +48,29 @@ class UploadingJobsDisplay extends Component {
...
+        return false;
+      };
+    } else {
+      window.onbeforeunload = null;
```
VictorSoroka1: This reset is also needed in `componentWillUnmount`, as this feature should only work on this page. Please create the method for this like `resetLeaveThePagePrevention`.

```suggestion
  resetLeaveThePagePrevention() {
    window.onbeforeunload = null;
  }
```

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
oleksii-popov: It should be OK as download process will not be interrupted by a browser.

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
oleksii-popov: Updated

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
oleksii-popov: Thanks, refactored

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
oleksii-popov: Done

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -48,8 +48,35 @@ class UploadingJobsDisplay extends Component {
...
   componentWillUnmount() {
     this.cancelFileRemovals();
+    this.resetLeaveThePagePrevention();
+  }
+
+  onLeaveHandler() {
+    const notAbleToLeave = Object.keys(this.props.files).some((file) => {
+      return this.props.files[file].status.toLowerCase() === 'uploading';
+    });
+    // prevent from leaving the page in case of download in progress
```
VictorSoroka1: Thanks for refactoring. Not a big deal, but please add empty row above the comment and remove the empty row below it so it would be more clear what comment refers to.

oleksii-popov: Thank, updated

PR #46, opened 2018-12-19
UIDATIMP-80: Handle inconsistent behavior when blocking file uploads …

VictorSoroka1:
…with different file extensions

### DIFF
--- a/src/components/ImportJobs/ImportJobs.js
+++ b/src/components/ImportJobs/ImportJobs.js
@@ -2,7 +2,6 @@ import React, { Component } from 'react';
 import { withRouter, Redirect } from 'react-router';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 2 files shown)


src/components/ImportJobs/ImportJobs.js
```diff
@@ -56,9 +55,11 @@ class ImportJobs extends Component {
    * @param  {Array<File>} files
    */
   validateFileExtensions(files = []) {
-    const filesType = get(files, [0, 'type'], '');
+    const fileTypeRegex = /\.(\w+)$/;
+    const filesType = files.map(({ name }) => name.match(fileTypeRegex)[1]);
```
OleksandrYehorovEPAM: `filesTypes` name seems to be more appropriate as it is an array of types and not a single type. Please consider renaming.
```suggestion
    const filesTypes = files.map(({ name }) => name.match(fileTypeRegex)[1]);
```

PR #68, opened 2019-01-15
UIDATIMP-97: Handle file name issue with file upload definition

VictorSoroka1:


src/components/UploadingJobsDisplay/utils/upload.js
```diff
@@ -91,33 +82,38 @@ export const uploadFile = (file, url, headers, onprogress) => {
...
-    const queryParams = `?fileId=${file.id}&uploadDefinitionId=${file.uploadDefinitionId}`;
-    const urlWithQueryParams = url + queryParams;
-
-    promise = promise
-      .then(() => uploadFile(file, urlWithQueryParams, headers, onXHRprogress))
-      .then(onXHRload)
-      .catch(onXHRerror);
-  });
+  Object
+    .keys(filesMeta)
```
OleksandrYehorovEPAM: Is there a reason to use `Object.keys()`? `Object.values()` seems more appropriate.

VictorSoroka1: I would leave `keys` approach for consistency as everywhere I use this approach. That includes rendering the files and other operations like this one.

PR #91, opened 2019-02-01
UIDATIMP-106: Prohibit file upload when there is a draft job in progress

OleksandrYehorovEPAM:
## Purpose
Prohibit file upload when there is a draft job in progress.

## Approach
Check if there is a draft job in progress before starting file upload. Show "File upload already in progress" message if there is a draft job in progress.

## Screenshots
screen shot 2019-02-01 at 13 41 57


src/components/ImportJobs/components/ReturnToAssignJobs/ReturnToAssignJobs.js
```diff
@@ -8,6 +9,8 @@ import { UploadingJobsContext } from '../../../UploadingJobsContextProvider';
 import css from './ReturnToAssignJobs.css';
 
 class ReturnToAssignJobs extends Component {
+  static propTypes = { prohibitFilesUploading: PropTypes.bool };
```
VictorSoroka1: Please, add default prop if it is not required.

src/components/ImportJobs/ImportJobs.js
```diff
@@ -53,23 +54,34 @@ class ImportJobs extends Component {
...
+      setFiles,
+      updateUploadDefinition,
+    } = this.context;
+
     this.setState({ isDropZoneActive: false });
 
-    const { setFiles } = this.context;
+    await updateUploadDefinition();
+
+    const { uploadDefinition } = this.context;
```
VictorSoroka1: Add this in the assignment on top.

src/components/ImportJobs/ImportJobs.js
```diff
@@ -133,8 +146,8 @@ class ImportJobs extends Component {
       >
         {openFileUploadDialogWindow => (
           <InvalidFilesModal
-            isModalOpen={isModalOpen}
-            onConfirmModal={this.hideModal}
+            isModalOpen={filesExtensionsModalOpen}
```
VictorSoroka1: Let's rename this prop to be just `isOpen`. The name of the component already makes it clear that the prop relates to the `modal`.

src/components/ImportJobs/ImportJobs.js
```diff
@@ -133,8 +146,8 @@ class ImportJobs extends Component {
...
+            onConfirmModal={this.hideFilesExtensionsModal}
```
VictorSoroka1: Let's rename this prop to be just `onConfirm`. The name of the component already makes it clear that the prop relates to the modal.

src/components/ImportJobs/ImportJobs.js
```diff
@@ -133,8 +146,8 @@ class ImportJobs extends Component {
...
             openFileUploadDialogWindow={openFileUploadDialogWindow}
```
VictorSoroka1: Let's rename this to be just `openDialogWindow`.

src/components/ImportJobs/components/ReturnToAssignJobs/ReturnToAssignJobs.js
```diff
@@ -24,11 +27,16 @@ class ReturnToAssignJobs extends Component {
 
   render() {
     const { deleteUploadDefinition } = this.context;
+    const { prohibitFilesUploading } = this.props;
+
+    const messageId = prohibitFilesUploading
```
VictorSoroka1: I think it can be used only one line for this statement instead of three.

```suggestion
    const messageId = prohibitFilesUploading ? 'messageWhenProhibited' : 'message';
```

src/components/ImportJobs/ImportJobs.js
```diff
@@ -53,23 +54,34 @@ class ImportJobs extends Component {
...
   onDrop = async acceptedFiles => {
+    const {
+      setFiles,
+      updateUploadDefinition,
+    } = this.context;
+
     this.setState({ isDropZoneActive: false });
 
-    const { setFiles } = this.context;
+    await updateUploadDefinition();
```
VictorSoroka1: Handle error cases. How the application will behave in case of network error?

src/components/ImportJobs/ImportJobs.js
OleksandrYehorovEPAM: This line must be after `await updateUploadDefinition()` as it fetches upladDefinition and writes it in context.

VictorSoroka1: Got it. Let's make `updateUploadDefinition` to return the definition so it is not needed to read for it from context. Also, it will make easier dealing with errors more clearly.

PR #177, opened 2019-04-02
UIDATIMP-132: add job profile form

OleksandrYehorovEPAM:


karma.conf.js
```diff
@@ -0,0 +1,2 @@
+/** @param {import('karma').Config} config */
+module.exports = config => config.set({ client: { captureConsole: false } });
```
VictorSoroka1: Is it needed?

OleksandrYehorovEPAM: It just makes the terminal output clearer when running tests.

src/components/JobProfilesForm/JobProfilesForm.js
```diff
@@ -0,0 +1,115 @@
...
+    pristine,
+    submitting,
+    handleSubmit,
+    onCancel,
+  } = props;
+
+  const isSubmitDisabled = pristine || submitting;
+
+  const paneTitle = <FormattedMessage id="ui-data-import.settings.jobProfiles.newJob" />;
+  const headLine = <FormattedMessage id="ui-data-import.settings.jobProfiles.newJob" />;
```
VictorSoroka1: Let's make it `const headLine = paneTitle;` since both are the same.

PR #178, opened 2019-04-03
UIDATIMP-6: Add icon for tags column

VictorSoroka1:


src/settings/JobProfiles/resultsFormatter.js
```diff
@@ -36,11 +43,20 @@ export const resultsFormatter = {
...
     const tags = get(record, 'tags.tagList', []);
 
-    if (!tags.length) {
+    if (isEmpty(tags)) {
       return '-';
     }
 
-    return tags.join(', ');
+    return (
+      <span className={css.tags}>
```
OleksandrYehorovEPAM: Looks weird for me to use `span` with `display: flex`. Consider using `div` or `display: inline-flex` if it will not break the markup

VictorSoroka1: Agree on using `display: inline-flex`

PR #181, opened 2019-04-05
UIDATIMP-164: Implement search job profiles feature

VictorSoroka1:


src/components/UploadingJobsDisplay/components/FileItem/FileItem.js
```diff
@@ -10,93 +10,91 @@ import { getFileItemMeta } from './getFileItemMeta';
...
-
-  deleteFile = () => {
+export const FileItem = memo(props => {
+  const deleteFile = () => {
     const {
       uiKey,
       status,
       onDelete,
-    } = this.props;
+    } = props;
```
OleksandrYehorovEPAM: Consider destructuring all props at the top of the component. Such an approach will make functions shorter.
```suggestion
    const deleteFile = () => onDelete(uiKey, status);
```

VictorSoroka1: Makes sense. Done.

PR #183, opened 2019-04-10
UIDATIMP-136: implement save job profile functionality

OleksandrYehorovEPAM:


translations/ui-data-import/en.json
```diff
@@ -108,5 +110,6 @@
   "validation.enterValue": "Please enter a value",
   "validation.fileExtension": "Please enter a valid file extension (e.g. .marc)",
   "validation.fileExtension.extension.invalid": "New record not created: File extension {value} is not a valid format",
-  "validation.fileExtension.duplication.invalid": "New record not created: File extension {value} already exists"
+  "validation.fileExtension.duplication.invalid": "New record not created: File extension {value} already exists",
```
VictorSoroka1: I thought about improving translations in order to reduce copy paste. I see that this translation string is the same as next one apart from entity type: 'File extension' versus 'Job Profile'. Please scan for similar translation strings and make them reusable.

src/settings/FileExtensions/FileExtensions.js
```diff
@@ -333,48 +246,42 @@ export class FileExtensions extends Component {
...
-              />
-            </div>
-            <div
-              className={sharedCss.fullWidthAndHeightContainer}
-              ref={this.createFullWidthContainerRef}
+          <div
+            className={sharedCss.container}
+            data-test-file-extensions
+          >
+            <SettingPage
```
VictorSoroka1: The approach is worthy. I think it makes sense to improve it by render prop pattern so the prop drilling would be avoided.
```
  <SettingsPage>
    {(props) => { // props needed for searchAndSort
       return (
         <Fragment>
           <SearchAndSort
             {...searchAndSortProps} // searchAndSortProps now should not be part of `SettingsPage` component and thus not be presented in its prop types
           >
           </SearchAndSort>
           ...
         <Fragment>
       )
    }}
  </SettingsPage>
```

PR #197, opened 2019-05-02
Fix unreadable error message when duplicating job profile

SergiyVSergiyenko:
When duplicating a job profile and changing only data type, an human unreadable error message appears.

### DIFF
--- a/src/components/SearchAndSort/SearchAndSort.js
+++ b/src/components/SearchAndSort/SearchAndSort.js
@@ -635,6 +635,8 @@ export class SearchAndSort extends Component {
       handleEditSuccess,
     } = this.props;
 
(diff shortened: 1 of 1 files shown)


VictorSoroka1:
@SergiyVSergiyenko, Please squash commits to only one.

src/components/SearchAndSort/SearchAndSort.js
```diff
@@ -635,6 +635,8 @@ export class SearchAndSort extends Component {
       handleEditSuccess,
     } = this.props;
 
+    const duplicateRecordInitialValues = { ...editRecordInitialValues, id: '' };
```
VictorSoroka1: Let's use a method from `lodash`

`omit(editRecordInitialValues, 'metadata', 'userInfo'),`

PR #203, opened 2019-05-03
UIDATIMP-189: Add status check for failed backend responses on the UI

VictorSoroka1:


src/utils/upload.js
```diff
@@ -0,0 +1,122 @@
+import { FILE_STATUSES, UPLOAD_DEFINTION_STATUSES } from './constants';
```
VictorSoroka1: TODO: format constants

PR #220, opened 2019-05-16
UIDATIMP-186: Implement landing page changes when temporary MARC Bib Load option button pushed

VictorSoroka1:


.eslintrc
```diff
@@ -51,6 +51,10 @@
     ],
     "semi": "off",
     "babel/semi": "error",
+    "object-curly-newline": ["error", {
```
OleksandrYehorovEPAM:
```suggestion
    "object-curly-newline": [
      "error",
      {
        "multiline": true,
        "minProperties": 2
      }
    ],
```

VictorSoroka1: Agree, changed.

PR #222, opened 2019-05-17
UIDATIMP-200: Reorganize tests folder structure and do several code improvements

VictorSoroka1:
## Purpose
Reorganize tests folder structure and do several code improvements [story](https://issues.folio.org/browse/UIDATIMP-200).

1) The tests folder structure is a bit messy and is not grouped by a feature which makes it hard to find the tests for a specific feature. As the testing files amount grow, eventually it will become a problem, so in the task, the test files should be grouped by topic.
2) The other improvement is to get rid of needless wrappers for FormattedMessage component as it the wrapped tag name can be passed to it.
3) Modify the existing eslint rule for the allowed amount of empty lines the end lines for consistency. Should be 1.
4) Replace the Log button with filename link on the logs list on the landing page


test/bigtest/tests/job-executions/job-item-test.js
```diff
@@ -22,7 +21,7 @@ describe('Job item', () => {
     });
 
     it('has correct date label', () => {
-      expect(jobsPane.previewJobs.jobItems(0).dateLabelText).to.be.equal(translation.endedRunning);
+      expect(jobsPane.previewJobs.jobItems(0).dateLabelText).contains('Ended');
```
YevheniiMaltsev: It looks like the developers will have to spend additional resources to support these values in case of some changes in translations. What do you think about it?

VictorSoroka1: It is okay in that case I think. This translation won't change in the near future.

PR #224, opened 2019-05-20
UIDATIMP-139: Save the new match profile

SergiyVSergiyenko:
Success_1

### DIFF
--- a/src/settings/MatchProfiles/MatchProfiles.js
+++ b/src/settings/MatchProfiles/MatchProfiles.js
@@ -154,6 +154,10 @@ export class MatchProfiles extends Component {
   defaultNewRecordInitialValues = {
     name: '',
     description: '',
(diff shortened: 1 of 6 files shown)


src/settings/MatchProfiles/MatchProfiles.js
```diff
@@ -154,6 +154,9 @@ export class MatchProfiles extends Component {
   defaultNewRecordInitialValues = {
     name: '',
     description: '',
+    incomingRecordType: 'MARC',
```
VictorSoroka1: Please add some TODO comment which says that these values are hardcoded now and will need to be changed in future.

taras-tkachenko: Will be moved to mockups untill backend will be ready to serve these for us.

PR #230, opened 2019-05-23
UIDATIMP-184: make job profile name a hotlink in match profile details

OleksandrYehorovEPAM:


src/settings/JobProfiles/resultsFormatter.js
```diff
@@ -28,21 +36,39 @@ export const resultsFormatter = (searchTerm, selectRecord, selectedRecords) => (
...
-      size="small"
-      app="data-import"
-      iconKey="jobProfiles"
-      className={sharedCss.cellAppIcon}
-    >
-      <HighLight
-        search={searchTerm}
-        className={sharedCss.container}
+  name: record => {
+    const name = (
```
VictorSoroka1: Since there is no one for `nameLink` button, I do not see a need in the variable here.

PR #248, opened 2019-06-20
UIDATIMP-144 and UIDATIMP-145: implement create new action profile fe…

OleksandrYehorovEPAM:
…ature

### DIFF
--- a/src/components/ActionProfilesForm/ActionProfilesForm.js
+++ b/src/components/ActionProfilesForm/ActionProfilesForm.js
@@ -1,0 +1,105 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import { Field } from 'redux-form';
+
+import {
+  Headline,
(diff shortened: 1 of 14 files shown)


src/components/ActionProfilesForm/ActionProfilesForm.js
```diff
@@ -0,0 +1,106 @@
...
+    pristine,
+    submitting,
+    handleSubmit,
+    onCancel,
+  } = props;
+
+  const isSubmitDisabled = pristine || submitting;
+
+  const paneTitle = <FormattedMessage id="ui-data-import.settings.actionProfiles.new" />;
+  const headLine = <FormattedMessage id="ui-data-import.settings.actionProfiles.new" />;
```
VictorSoroka1: I would just set it to equal `paneTitle` as it is the same.

OleksandrYehorovEPAM: Done

PR #257, opened 2019-07-03
Update action types icons

rasmuswoelk:
## Changes
- Updated action type icons to use the correct icons from stripes-components
- Removed local icons for "combine", "create" and "replace".

## Before
image

## After
image


OleksandrYehorovEPAM:
@rasmuswoelk thank you for these changes. Seems like `iconStyle="action"` can also be applied to the modify icon to have the same style for all action icons. I see then `Icon` component breaks margins between the icon and its label when the label takes more than one line but don't worry I will take care of this.

PR #288, opened 2019-08-14
UIDATIMP-251: Implement Match Profile structure changes :: DONE

taras-tkachenko:
It was required to take Existing Field definition from MatchDetails[].fields object and treat them not as single field value but as a path with subfields.


src/components/ListTemplate/ColumnTemplates/MatchColumn.js
```diff
@@ -27,14 +27,19 @@ export const MatchColumn = memo(({
...
   const fieldSource = (field || existingRecordType || '').replace(/_/g, ' ');
-  const fieldMatched = (fieldMarc || fieldNonMarc || existingStaticValueType || '').replace(/_/g, ' ');
+  const matchRule = matchDetails[0];
+  const fieldsMatched = matchRule.existingMatchExpression.fields.flatMap(item => item.value);
+
+  if (document.dir === HTML_LANG_DIRECTIONS.RIGHT_TO_LEFT) {
+    fieldsMatched.reverse();
+  }
+
+  const fieldMatched = fieldsMatched.join(fieldsMatched.length > 1 ? '.' : '').replace(/_/g, ' ');
```
VictorSoroka1:
```suggestion
  const fieldMatched = fieldsMatched.join('.').replace(/_/g, ' ');
```
I think it can be simplified.

src/components/ListTemplate/ColumnTemplates/MatchColumn.js
```diff
@@ -27,14 +27,19 @@ export const MatchColumn = memo(({
...
-    existingStaticValueType,
     field,
-    fieldMarc,
-    fieldNonMarc,
+    matchDetails,
   } = record;
 
   const fieldSource = (field || existingRecordType || '').replace(/_/g, ' ');
-  const fieldMatched = (fieldMarc || fieldNonMarc || existingStaticValueType || '').replace(/_/g, ' ');
+  const matchRule = matchDetails[0];
```
OleksandrYehorovEPAM: I don't think this variable is really needed.

src/components/ListTemplate/ColumnTemplates/MatchColumn.js
```diff
@@ -27,14 +27,19 @@ export const MatchColumn = memo(({
...
+  const fieldsMatched = matchRule.existingMatchExpression.fields.flatMap(item => item.value);
```
OleksandrYehorovEPAM: Many of those fields are not required in the schema. It means that they could be empty. Please use lodash.get function to avoid errors.

OleksandrYehorovEPAM: Please use map

src/components/ListTemplate/ColumnTemplates/MatchColumn.js
taras-tkachenko: Done

src/components/ListTemplate/ColumnTemplates/MatchColumn.js
taras-tkachenko: Done

src/components/ListTemplate/ColumnTemplates/MatchColumn.js
taras-tkachenko: Done

taras-tkachenko: Done

PR #290, opened 2019-08-20
UIDATIMP-257, UIDATIMP-258 and UIDATIMP-252

OleksandrYehorovEPAM:


src/utils/getFileExtension.js
```diff
@@ -0,0 +1,11 @@
+/** @param {File | { name: string }} file */
+export const getFileExtension = file => {
+  if (file && file.name) {
+    const fileExtensionRegExp = /\.(\w+)$/;
+    const [extension] = (file.name.match(fileExtensionRegExp) || []);
+
+    return extension.toLowerCase();
```
VictorSoroka1: `const [extension] = (file.name.match(fileExtensionRegExp) || []);`

Theoretically `extension` can become `undefined` in case match ended up to be `null`. In that case, calling `toLowerCase` would end up in error.

src/components/ListView/ListView.js
```diff
@@ -180,17 +183,24 @@ export class ListView extends Component {
...
       RecordForm,
       renderHeaders,
+      actionMenuItems,
     } = this.props;
     const { showRestoreModal } = this.state;
 
     const urlQuery = queryString.parse(search);
     const searchTerm = trimSearchTerm(urlQuery.query);
 
+    const actionMenu = Array.isArray(actionMenuItems) && !!actionMenuItems.length
```
VictorSoroka1: I see this chunk repeating in other profiles. I am thinking about simplifying it or more to utils. I am thinking about 
`!isEmpty(actionMenuItems)` util from `lodash`. Please check if it does make sense,

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -100,10 +107,10 @@ export class UploadingJobsDisplay extends Component {
   mapFilesToState() {
     const {
       history,
-      location: { state = {} },
+      location,
     } = this.props;
 
-    const { files } = state;
+    const { state: { files } = {} } = location;
```
VictorSoroka1: Since `get` util is already used in the file I think it makes sense to simplify this code.

`const files = get(location, 'state.files')`

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -141,7 +151,9 @@ export class UploadingJobsDisplay extends Component {
   };
 
   handleNavigation = nextLocation => {
-    const shouldPrompt = this.filesUploading;
+    const { location } = this.props;
```
VictorSoroka1: Is it still the rule about placing the space after this.props?

src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -364,6 +376,12 @@ export class UploadingJobsDisplay extends Component {
 
       console.error(error); // eslint-disable-line no-console
     }
+
+    const lastFileDeleted = Object.values(get(this.state, ['files'], {})).length === 0;
```
VictorSoroka1:
```suggestion
    const lastFileDeleted = isEmpty(get(this.state, ['files'], {}));
```

PR #291, opened 2019-08-21
UIDATIMP-248 & UIDATIMP-249 :: FIXED

taras-tkachenko:
UIDATIMP-248: INTL translation messages missing renewed.
UIDATIMP-249: Callout appearance removed on HTTP :: 409 status errors.


src/components/ViewContainer/getCRUDActions.js
```diff
@@ -36,7 +36,9 @@ export const getCRUDActions = ({
       return response || { ok: true };
     })
     .catch(error => {
-      onError('delete', record, error);
+      if (error.status !== 409) {
```
VictorSoroka1: I think it makes sense to move this check to a variable to add more context on this.

taras-tkachenko: It's an exceptional case for just one error types (associations validation fails) so I cannot see how to move it to a variable and where to place it aside of here.

PR #292, opened 2019-08-22
UIDATIMP-252: fix "Load MARC records" button

OleksandrYehorovEPAM:


src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
```diff
@@ -531,11 +530,11 @@ export class UploadingJobsDisplay extends Component {
   // the process of loading MARC BIB files (UIDATIMP-185)
   renderLoadMarcButton = menu => {
     const {
-      files = [],
+      files = {},
       recordsLoadingInProgress,
     } = this.state;
 
-    const areMarcFiles = every(files, file => file.name.match(/\.(marc|mrc)$/, 'i'));
+    const areMarcFiles = Object.values(files).every(file => file.name.match(/\.(marc|mrc)$/i));
```
VictorSoroka1:
```suggestion
    const areMarcFiles = every(files, file => file.name.match(/\.(marc|mrc)$/i));
```
Just for curiosity, why did you decide to get rid of `every` from `lodash`? I found it is handy for cases like that in order to avoid low-level Object.values.

PR #299, opened 2019-09-10
UIDATIMP-217: add rowUpdater to fix checkboxes

OleksandrYehorovEPAM:


src/components/ListView/ListView.js
```diff
@@ -165,6 +165,12 @@ export class ListView extends Component {
     />
   );
 
+  rowUpdater = ({ id }) => {
+    const { checkboxList: { selectedRecords } } = this.props;
+
+    return selectedRecords && selectedRecords.has(id);
```
VictorSoroka1: I see that for similar cases below you have used just 
`return selectedRecords.has(id);`

Is there additional check is needed?

OleksandrYehorovEPAM: The check is needed because `checkboxList` is not a required property. However, I am going to refactor it.

VictorSoroka1: Sounds good.

VictorSoroka1:
@OleksandrYehorovEPAM, please also add a note to the changelog file.

PR #304, opened 2019-09-19
UIDATIMP-278: Create Section Component

mariia-aloshyna:


src/components/Section/Section.js
```diff
@@ -0,0 +1,61 @@
...
+          ? optionalHeader(caption)
+          : (<h3 className={css.heading}>{caption}</h3>)
+        }
+      </header>
+    );
+  };
+
+  return (
+    <div className={css.container}>
+      {renderHeader(headerCaption)}
```
OleksandrYehorovEPAM: Please rewrite functions to use values from the scope and not from params

src/components/Section/Section.js
```diff
@@ -0,0 +1,61 @@
...
+  const [isChecked, setChecked] = useState(true);
+
+  const optionalHeader = caption => {
+    return (
+      <div>
+        <label
+          className={css.heading}
+          htmlFor={sectionId}
+        >
+          <input
```
OleksandrYehorovEPAM: Please use `Checkbox` component from `stripes-components`

src/components/Section/Section.js
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <div className={css.container}>
+      {renderHeader(headerCaption)}
+      {isChecked && children}
+    </div>
+  );
+});
+
+Section.propTypes = {
+  headerCaption: PropTypes.string.isRequired,
```
OleksandrYehorovEPAM: This prop should be of type `PropTypes.node.isRequired` to allow usage of `<FormattedMessage>`

src/components/Section/Section.js
```diff
@@ -0,0 +1,61 @@
...
+  isOptional: PropTypes.bool,
+  sectionId: PropTypes.string,
```
OleksandrYehorovEPAM: Please consider renaming it it just `id`

src/components/Section/Section.css
```diff
@@ -0,0 +1,23 @@
+@import "@folio/stripes-components/lib/variables.css";
+
+.container {
+  border-radius: 1rem;
+  border: 1px solid var(--color-border);
+  background-color: var(--color-fill);
+  padding: 20px;
+}
+
+.container :nth-child(n+2) {
```
OleksandrYehorovEPAM: Can we avoid such a selector?

src/components/Section/Section.js
```diff
@@ -0,0 +1,61 @@
...
+          />
+          {caption}
+        </label>
+      </div>
+    );
+  };
+
+  const renderHeader = caption => {
+    return (
+      <header>
```
OleksandrYehorovEPAM: Consider using `Headline` component from `stripes-components`

src/components/Section/Section.js
```diff
@@ -0,0 +1,55 @@
...
+  headerCaption,
+  isOptional,
+  id,
+  children,
+}) => {
+  const [isChecked, setChecked] = useState(true);
+
+  const headline = (
+    <Headline
+      size="1.17rem"
```
OleksandrYehorovEPAM: Please use one of the available size values `size: PropTypes.oneOf(['small', 'medium', 'large', 'x-large', 'xx-large'])`

src/components/Section/Section.css
```diff
@@ -0,0 +1,12 @@
+@import "@folio/stripes-components/lib/variables.css";
+
+.container {
+  border-radius: 1rem;
+  border: 1px solid var(--color-border);
+  background-color: var(--color-fill);
+  padding: 20px;
```
OleksandrYehorovEPAM: Please use `rem`-s

src/components/Section/Section.css
```diff
@@ -0,0 +1,12 @@
...
+}
+
+.container > :nth-child(n+2) {
+    margin-top: 20px;
```
OleksandrYehorovEPAM: Please use rem-s

src/components/Section/Section.js
```diff
@@ -0,0 +1,55 @@
...
+      fullWidth
+      label={headline}
+      id={id}
+      checked={isChecked}
+      onChange={() => setChecked(!isChecked)}
+    />
+  );
+
+  return (
+    <div className={css.container}>
```
OleksandrYehorovEPAM: Since it is a `Section` component probably `section` tag would be batter than `div`

src/components/Section/Section.js
```diff
@@ -0,0 +1,55 @@
...
+      {isOptional ? optionalHeadline : headline}
+      {isChecked && children}
+    </div>
+  );
+});
+
+Section.propTypes = {
+  headerCaption: PropTypes.node.isRequired,
+  isOptional: PropTypes.bool,
+  id: PropTypes.string,
```
OleksandrYehorovEPAM: Do we really need this prop?

src/components/Section/Section.js
mariia-aloshyna: Can I add a class with font-size property to this component? The thing is I'd like components to be similar in style (RecordTypesSelect has a headline with 1.17rem font-size value). If it doesn't matter I can replace a size value with a large value.

OleksandrYehorovEPAM: For now, just use one of the values provided by `Headline` component. I will check the styles in `RecordTypesSelect` component.

src/components/Section/Section.js
taras-tkachenko: No.
We can add this as HTML prop if needed

taras-tkachenko:
Code is OK now, waiting for tests.

PR #307, opened 2019-09-24
UIDATIMP-260: Add the record totals in the log list and shrink the size of panes

mariia-aloshyna:


test/bigtest/mocks/jobs-logs.js
```diff
@@ -4,6 +4,7 @@ const jobsLogsData = [
     jobExecutionHrId: '02',
     jobProfileName: 'Multilingual support check',
     fileName: 'import_28.mrc',
+    totalRecords: 9.822,
```
OleksandrYehorovEPAM: Please make it an integer. `totalRecords` is an integer in the schema https://github.com/folio-org/data-import-raml-storage/blob/0f54dcefdcd3bba44e55a243dc05e6d92d356d4d/schemas/dto/logDto.json#L25

PR #310, opened 2019-09-26
UIDATIMP-207: add details section on create/edit action profile form

OleksandrYehorovEPAM:


src/settings/ActionProfiles/ActionProfilesForm.js
```diff
@@ -22,20 +38,96 @@ import {
...
+          ACTION_PROFILES_FORM_FOLIO_RECORD_TYPES.MARC_HOLDINGS.type,
+        ]);
+      }
+      case ACTION_TYPES.REPLACE.type: {
+        return omit(ACTION_PROFILES_FORM_FOLIO_RECORD_TYPES, [
+          ACTION_PROFILES_FORM_FOLIO_RECORD_TYPES.ORDER.type,
+          ACTION_PROFILES_FORM_FOLIO_RECORD_TYPES.INVOICE.type,
+        ]);
+      }
+      case ACTION_TYPES.CREATE.type:
```
VictorSoroka1: Not a big deal, but this line is excess really since it is handled by the default case.

OleksandrYehorovEPAM: I agree but I did it intentionally for readability purposes. Here and in the function above.

PR #320, opened 2019-10-10
UIDATIMP-274: Create Find Import Profiles Plugin: Create Profile Shapes :: DONE

taras-tkachenko:


src/settings/JobProfiles/JobProfiles.js
```diff
@@ -26,6 +26,85 @@ import { JobProfilesForm } from './JobProfilesForm';
...
+  renderHeaders: props => {
+    let headers = {
+      name: <FormattedMessage id="ui-data-import.name" />,
+      description: <FormattedMessage id="ui-data-import.description" />,
+      tags: <FormattedMessage id="ui-data-import.tags" />,
+      updated: <FormattedMessage id="ui-data-import.updated" />,
+      updatedBy: <FormattedMessage id="ui-data-import.updatedBy" />,
+    };
+
+    if (props && props.checkboxList) {
```
VictorSoroka1: I would use default assigning like `renderHeaders: (props = {}) => {`, so no additional checks for it is needed. The is also true for [here](https://github.com/folio-org/ui-data-import/pull/320/files#diff-f006c40166676f423d25786d486ab283R95), [here](https://github.com/folio-org/ui-data-import/pull/320/files#diff-e7ac3d689217081429bf909a5ddae36fR92) and [here](https://github.com/folio-org/ui-data-import/pull/320/files#diff-6da8b35c659aa25cd4ef88ca4e5f2cf4R90-R96).

src/settings/JobProfiles/JobProfiles.js
```diff
@@ -26,6 +26,85 @@ import { JobProfilesForm } from './JobProfilesForm';
...
+  visibleColumns: [
+    'name',
+    'description',
+    'tags',
+    'updated',
+    'updatedBy',
+  ],
+  columnWidths: {
+    isChecked: '35px',
+    name: '300px',
```
VictorSoroka1: Are hardcoded widths still needed? MCL already has already a mechanism for calculating widths automatically, so `columnWidths` usage can be reduced. The same is true for other profiles.

PR #363, opened 2019-11-28
UIDATIMP-325: Create Match Profiles Form renderer

mariia-aloshyna:


src/components/FlexibleForm/Control.js
```diff
@@ -139,23 +197,26 @@ export const Control = memo(props => {
...
+  staticControlType: PropTypes.string.isRequired,
   component: PropTypes.string,
   label: PropTypes.string || Node,
   intl: intlShape,
   styles: PropTypes.shape(PropTypes.string),
   classNames: PropTypes.arrayOf(PropTypes.string),
   dataOptions: PropTypes.arrayOf(PropTypes.shape(PropTypes.string)),
   repeatable: PropTypes.bool,
-  children: PropTypes.arrayOf(Node),
+  childrenControl: PropTypes.arrayOf(Node),
```
taras-tkachenko: Why to change names?
children is much shorter...

mariia-aloshyna: I changed this prop name because React Component has own "children" prop, and this prop is used in the Headline component, so when we render control children it makes confuse

PR #410, opened 2020-01-22
UIDATIMP-317: Add profile type to name and set background

mariia-aloshyna:


src/components/ListTemplate/ColumnTemplates/DefaultColumn.js
```diff
@@ -22,10 +23,10 @@ export const DefaultColumn = memo(({
         search={searchTerm || ''}
         className={sharedCss.container}
       >
-        {value}
+        {customValue || value}
       </HighLight>
     )
-    : value;
+    : customValue || value;
```
Vladyslav-Velytskyi: Please cover it into another variable and use it in 26 string too

PR #424, opened 2020-02-10
UIDATIMP-314: Prohibit adding match profile after action profile

mariia-aloshyna:
## Description
Additional Match profiles can not be attached after Action profile

## Approach
Add `siblingsProhibited` property to the `PROFILE_LINKING_RULES` constant. Disable the `LinkerButton` for a prohibited profile.

## Issue
[UIDATIMP-314](https://issues.folio.org/browse/UIDATIMP-314)

## Screenshot
Disabled_profile


src/components/ProfileTree/ProfileLinker/LinkerMenu.js
```diff
@@ -53,6 +60,7 @@ LinkerMenu.propTypes = {
   onToggle: PropTypes.func.isRequired,
   onClick: PropTypes.func.isRequired,
   keyHandler: PropTypes.func,
+  disabledOptions: PropTypes.arrayOf(PropTypes.string),
```
Vladyslav-Velytskyi: Shouldn't this value be required

mariia-aloshyna: I guess this prop is not required, added this to defaultProps.

PR #458, opened 2020-03-12
UIPFIMP-11: Support static value submatches

mariia-aloshyna:
## Purpose
It is needed to change the query string depending on the associated match profile position in the job profile tree.
## Approach

- Check match profile parent type:
  - if exists - position of the match profile is secondary, tertiary, etc;
  - if not - position of the match profile is primary;
- Determine the query value depending on the position and pass it to the plugin
- Build a query string in the manifest depending on the query value passed to the plugin

## Issue
[UIPFIMP-11](https://issues.folio.org/browse/UIPFIMP-11)


src/settings/MatchProfiles/MatchProfiles.js
```diff
@@ -98,21 +107,15 @@ export const matchProfilesShape = {
...
-            },
-            [],
-          ),
-        },
-        staticFallback: { params: {} },
+      params: (_q, _p, _r, _l, props) => {
+        const findAll = 'cql.allRecords=1';
+        const sort = get(_r, ['query', 'sort'], null);
+        const search = get(_r, ['query', 'query'], null);
+        const sortQuery = `${sort ? `sortBy ${getSortQuery(sortMap, sort)}` : ''}`;
```
Vladyslav-Velytskyi: bit hard coded, it might be easier
```suggestion
        const sortQuery = sort ? `sortBy ${getSortQuery(sortMap, sort)}` : '';
```

mariia-aloshyna: Forgot to change it, thanks!

PR #493, opened 2020-04-05
UIDATIMP-445: Render Repeatable Fields as MultiColumnList in FlexibleForm static view when needed :: DONE

taras-tkachenko:


src/components/FlexibleForm/Control.js
```diff
@@ -219,37 +219,111 @@ export const Control = memo(props => {
...
+        <div>
+          <strong>{isTranslationId(legend) ? <FormattedMessage id={legend} /> : legend}</strong>
+        </div>
+        <Cmp
+          contentData={data}
+          visibleColumns={visibleColumns}
+          columnWidths={columnWidths}
+          columnMapping={mapping}
+        />
+        <br />
```
mariia-aloshyna, line 291: Why is the `<br>` tag needed here?

taras-tkachenko: It's needed to simply separate MCL from the next section

src/components/FlexibleForm/Control.js
```diff
@@ -219,37 +219,109 @@ export const Control = memo(props => {
...
-              referenceTables={referenceTables}
-              setReferenceTables={setReferenceTables}
-              initialFields={initialFields}
-              injectedProps={injectedProps}
-              commonSections={commonSections}
-              record={attribs.record}
-              {...cfg}
-            />
-          ))}
+          {repeatable ? refTable.map(() => renderChildren()) : renderChildren()}
```
VictorSoroka1, line 324: I see that for the `repeatable` the .map is used both for `refTable` and for [`children`](https://github.com/folio-org/ui-data-import/pull/493/files#diff-6709362c6d184124da81079fe4cef458R222). Please clarify whether it is correct.

In case there are multiple `refTable` and only one child the `repeatableIndex` for each of the child will be `0` because of that.

In case the `index`

src/components/FlexibleForm/Control.js
mariia-aloshyna: Isn't the MCL component block-level element?

src/components/FlexibleForm/Control.js
taras-tkachenko: Yes, all is correct

src/components/FlexibleForm/Control.js
taras-tkachenko: Yes, but it requires styles adjustments which is not wanted here

src/components/FlexibleForm/Control.js
taras-tkachenko: This is done for the cases when the field names and values are not flat an obvious as usual:
```
{
    name: 'notes',
    enabled: true,
    path: 'item.notes',
    value: '',
    subfields: [{
      order: 0,
      path: 'item.notes[]',
      fields: [{
        // order: 0,
        name: 'itemNoteTypeId',
        enabled: true,
        path: 'item.notes[].itemNoteTypeId',
        value: '',
      }, {
        // order: 1,
        name: 'note',
        enabled: true,
        path: 'item.notes[].note',
        value: '',
      }, {
        // order: 2,
        name: 'staffOnly',
        enabled: true,
        path: 'item.notes[].staffOnly',
        value: null,
        booleanFieldAction: 'AS_IS',
      }],
    }],
  }
```

Here reference table for the current repeatable is taken from `subfields` prop, but columns are in its `fields` subsection.

PR #502, opened 2020-04-09
UIDATIMP-477: Implement <ProhibitionIcon> component

mariia-aloshyna:
## Description
Implement a "value prohibited" component with "No value allowed" background wording.

## Ticket
[UIDATIMP-477](https://issues.folio.org/browse/UIDATIMP-477)

## Screenshot
valueProhibited


src/components/ProhibitionIcon/ProhibitionIcon.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import css from './ProhibitionIcon.css';
+
+export const ProhibitionIcon = () => (
+  <FormattedMessage id="ui-data-import.noValueAllowed">
```
taras-tkachenko, line 7: Please make the message configurable through the optional prop with the value above as default.
Let it accept `<FormattedMessage />` and translation ID both.

mariia-aloshyna: Done. I added optional prop 'ariaLabelId', which accepts translation ID. I didn't let it accept `<FormattedMessage />` because in both cases (ariaLabelId prop is passed or not) it will be rendered with `<FormattedMessage />` component but just IDs can be different.

PR #513, opened 2020-04-16
UIDATIMP-479: Optional sections are collapsed even if contain data :: FIXED

IhorKoba:
## Description
Optional sections are collapsed even if contain data

## Approach
1. Change logic for defining *isOpen* parameter, which is passing to Section.js from Control.js component in renderCommonSection to detect how to render content based on data availability
2. Add getChildrenWithName function to get only objects with name from form config and skip “layout” objects such as 'Col' and 'Row'

## Issue
[UIDATIMP-479](https://issues.folio.org/browse/UIDATIMP-479)


src/components/Section/Section.js
```diff
@@ -20,9 +20,10 @@ export const Section = memo(({
   isOpen,
   className,
   children,
+  isVisible = true,
   ...rest
 }) => {
-  const [isChecked, setChecked] = useState(optional ? isOpen : true);
+  const [isChecked, setChecked] = useState(optional ? isOpen : isVisible);
```
mariia-aloshyna, line 26: Doesn't `isOpen` prop have the same meaning as `isVisible`? Can we change the line [here](https://github.com/folio-org/ui-data-import/pull/513/files#diff-6709362c6d184124da81079fe4cef458R492) to `isOpen: hasData` e.g. and then `isVisible` prop will not be needed?

test/bigtest/mocks/form-config-samples.js
```diff
@@ -326,6 +326,7 @@ export const formConfigSamples = [{
     sectionKey: 'qualifierPart',
     staticControlType: 'Section',
     label: 'ui-data-import.match.**ns**.part',
+    name: '**ns**qualifierPart',
```
mariia-aloshyna, line 329:
```suggestion
    name: '**ns**.qualifierPart',
```

test/bigtest/tests/match-profiles/match-profile-details-test.js
```diff
@@ -200,11 +200,8 @@ describe('Match Profile View', () => {
               expect(matchProfileDetails.matchCriteria.incomingRecordSections.children(1).label).to.be.equal('Use a qualifier');
             });
 
-            it('has empty QualifierType field value', () => {
-              expect(matchProfileDetails.matchCriteria.fieldQualifierType.value.text).to.be.equal('-');
-            });
-            it('has empty QualifierValue field value', () => {
-              expect(matchProfileDetails.matchCriteria.fieldQualifierValue.value.text).to.be.equal('-');
```
mariia-aloshyna, line 207: I think it will be good to have a covered with tests case when the `Use a qualifier` section has one filled and one empty field.

test/bigtest/mocks/form-config-samples.js
IhorKoba: removed unnecessary name

test/bigtest/tests/match-profiles/match-profile-details-test.js
IhorKoba: will be added in the future

PR #524, opened 2020-04-24
UIDATIMP-407: Implement date picker decorator

mariia-aloshyna:
## Description
Implement date picker component as a regular components decorator.

## Approach
* Implement custom `<TextDate>` component similar to the stripes `<Datepicker>` component except it allows to enter any text to the input field
* Implement `withDatePicker `decorator. Check if selected wrapper value == `"CHOOSE_DATE"` then render the `<TextDate>` component otherwise render common `<TextField>` component
* Adjust `<OptionsList>` component

## Screen record
datepicker


src/components/FlexibleForm/ControlDecorators/withReferenceValues/withReferenceValues.js
```diff
@@ -119,7 +119,7 @@ export const withReferenceValues = memo(props => {
               id={id}
               label={localized}
               dataOptions={dataOptions}
-              optionValue="id"
+              optionValue="name"
```
taras-tkachenko, line 122: Why did you changed this to "name"?

src/components/FlexibleForm/ControlDecorators/withReferenceValues/withReferenceValues.js
```diff
@@ -132,7 +132,7 @@ export const withReferenceValues = memo(props => {
           id={id}
           label={wrapperLabel}
           dataOptions={dataOptions}
-          optionValue="id"
+          optionValue="name"
```
taras-tkachenko, line 135: The same as above

src/components/FlexibleForm/ControlDecorators/withReferenceValues/withReferenceValues.js
mariia-aloshyna: Because in the `OptionsList` component I changed function `onClick()`, now it passes` option?.[optionValue] || option?.[optionLabel]` to the `onSelect()` method. I did this because in the `withDatePicker` decorator we don't put options values to the input field as it is (moreover values in this decorator are rendered not as a string, but as an object), but we check the option name.
So due to changes in the `OptionsList` component, I had to change `optionValue` here.
It works well, tested it.

PR #535, opened 2020-05-05
UIDATIMP-478 Hotlinks has been added for match and action profiles fo…

IvanKryzhanovskyi:
## Description
view the details of those other profiles without having to manually search for them

## Approach

- property showAsHotLink added to appropriate Tree component in Job profile Overview section
- Logic for ListTemplate "Default column" render process has been updated. Check if column has property showAsHotLink 
- If yes - display a Button component with link to appropriate entity details view.
- If not - display as usual - AppIcon with plain text title

## Task
[UIDATIMP-478](https://issues.folio.org/browse/UIDATIMP-478)

## Screenshot
image


mariia-aloshyna:
I suggest creating another one list template (kind of `nameLink`) instead of extending the existing `name` with a flag (`showAsHotLink`). I'll try to explain - let's imagine that we will need to make the `name template` underlined, but the `linked name` should not be underlined, so we will have to pass some another prop-flag to the `name` template to distinguish it (this example is not quite real-life suitable, but I hope the logic is clear), so I think will be better to have two separate independent templates.

src/components/ListTemplate/ColumnTemplates/DefaultColumn.js
```diff
@@ -2,8 +2,12 @@ import React, { memo } from 'react';
 import HighLight from 'react-highlighter';
 import PropTypes from 'prop-types';
 
+import { Button } from '@folio/stripes/components';
+
```
mariia-aloshyna, line 6: Remove the blank line, please. This import and import below belong to one group - imports of stripes components

src/components/ListTemplate/ColumnTemplates/DefaultColumn.js
```diff
@@ -12,13 +16,17 @@ export const DefaultColumn = memo(({
   className = sharedCss.cellAppIcon,
   iconKey,
```
mariia-aloshyna, line 17: Lets replace `iconKey` prop with `entityKey` and then on the 45 line pass `entityKey` to the `<AppIcon>` component's `iconKey` prop. In this case, you don't need [this manipulation](https://github.com/folio-org/ui-data-import/pull/535/files#diff-c7438818fbd56fd424a8c77cebbe3a9bR28) and this help to avoid problems, because `iconKey` may differ from the entity name, as a result of which we can get an invalid link.
```suggestion
  entityKey,
```

src/components/ListTemplate/ColumnTemplates/DefaultColumn.js
```diff
@@ -44,6 +48,26 @@ export const DefaultColumn = memo(({
...
+  return (
+    <>
+      {showAsHotLink ? (
+        <Button
+          data-test-profile-link
+          buttonStyle="link"
+          marginBottom0
+          to={`/settings/data-import/${entityName}/view/${recordId}`}
+          buttonClass={sharedCss.cellLink}
+        >
```
mariia-aloshyna, line 65: Consider moving this to a separate variable
```suggestion
        {hotLink}
```

src/components/ProfileTree/ProfileTree.js
```diff
@@ -38,6 +38,7 @@ export const ProfileTree = memo(({
   record,
   className,
   dataAttributes,
+  showAsHotLink,
```
mariia-aloshyna, line 41: Just suggestion. For now, it looks like you want to show the whole tree as a hotlink :)
```suggestion
  showLabelsAsHotLink,
```

src/components/ProfileTree/ProfileTree.js
```diff
@@ -156,6 +157,7 @@ export const ProfileTree = memo(({
                 onUnlink={unlink}
                 onDelete={remove}
                 okapi={okapi}
+                showAsHotLink={showAsHotLink}
```
mariia-aloshyna, line 160: The same for the `ProfileBranch` component, for the `ProfileLabel` it can stay as it is

src/components/ListTemplate/ColumnTemplates/DefaultColumn.js
IvanKryzhanovskyi: new object in constants.js has been created for profile type mapping

test/bigtest/tests/job-profiles/job-profile-details-test.js
```diff
@@ -69,6 +70,19 @@ describe('Job Profile View', () => {
...
+    describe('when profile button is clicked', () => {
+      let linkProfileType;
+
+      beforeEach(async function () {
+        linkProfileType = jobProfileDetails.hotLink.$root.dataset.testProfileLink;
+        await jobProfileDetails.hotLink.click();
+      });
+
+      it('redirects to appropriate profile details page', () => {
+        expect(location().pathname).contains(linkProfileType);
```
mariia-aloshyna, line 82: I suggest a comparison with string here, it would be more readable
```suggestion
        expect(location().pathname).contains('/data-import/{linkProfileType}/view');
```

VictorSoroka1: What about `endsWith`?

test/bigtest/interactors/job-profiles/job-profile-details.js
```diff
@@ -23,6 +24,7 @@ import { ActionMenuInteractor } from '../action-menu-interactor';
   confirmationModal = new ConfirmationModalInteractor('#delete-job-profile-modal');
   runConfirmationModal = new ConfirmationModalInteractor('#run-job-profile-modal');
   callout = new CalloutInteractor();
+  hotLink = new Interactor('[class*="profile-tree---"] [data-test-profile-link]');
```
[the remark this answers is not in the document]
VictorSoroka1: Is there is a way to replace [class*="profile-tree---"] with data-test attribute?

test/bigtest/tests/job-profiles/job-profile-details-test.js
VictorSoroka1: Agree, it is better to compare with string as in case the `href` change to the incorrect one the test would still pass as the `location().pathname` reflects its value. In turn, comparing to the string representation would make sure that the link is correct.

IvanKryzhanovskyi:
https://issues.folio.org/browse/UIDATIMP-527 has been created to cover additional test cases

PR #538, opened 2020-05-12
UIDATIMP-517 update to react-intl v4

IvanKryzhanovskyi:
react-intl no longer uses the legacy `contextTypes` API, so we're back to `injectIntl`.

### DIFF
--- a/src/settings/JobProfiles/ViewJobProfile.js
+++ b/src/settings/JobProfiles/ViewJobProfile.js
@@ -31,6 +31,7 @@ import {
   stripesConnect,
   stripesShape,
 } from '@folio/stripes/core';
+import SafeHTMLMessage from '@folio/react-intl-safe-html';
 
 import {
   ENTITY_KEYS,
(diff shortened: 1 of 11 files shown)


src/settings/JobProfiles/ViewJobProfile.js
```diff
@@ -482,7 +480,7 @@ export class ViewJobProfile extends Component {
           open={this.state.showRunConfirmation}
           heading={<FormattedMessage id="ui-data-import.modal.jobProfile.run.header" />}
           message={(
-            <SafeHTMLMessage
+            <FormattedMessage
```
mariia-aloshyna, line 483: Why have you replaced `SafeHTMLMessage` with `FormattedMessage`?

mariia-aloshyna:
Update CHANGELOG, please.

src/settings/JobProfiles/ViewJobProfile.js
mariia-aloshyna: I see that using of translation values with attributes in translations are no longer supported, but you may still use `react-intl-safe-html` for cases of **b**, **i**, **em**, **strong**, **span**, and **div** _without  attributes_ - this is our case.

zburke:
I hope I am not stepping on any toes by merging this, but we needed to merge it in order to repair settings in platform-complete. Because of the way settings components are loaded, any one component that includes a build-time error -- such as importing `intlShape` when it doesn't exist -- completely breaks settings. 

Obviously, that's it's own problem that we need to address, but merging this update is the short way around it.

PR #540, opened 2020-05-13
Bump up stripes-acq-components version

aliaksei-chumakou:


aliaksei-chumakou:
@taras-tkachenko Hi Taras, maybe it makes sense to update lib version since it was released a while ago (Q1), thanks

PR #545, opened 2020-05-18
UIDATIMP-515: Fix Item status list in field mapping profiles

mariia-aloshyna:
## Description
In Item field mapping profiles, the reference dropdown list for Item status is wrong.

## Approach
Statuses are stored as ENUM on the server, so it's not needed to retrieve it from the server.

- Add `acceptedValuesList` prop to the `AcceptedValuesField` component
- Set `acceptedValuesList` to the `listOptions` if it's passed
- Check `wrapperSourceLink && wrapperSourcePath`. If it's not empty then fetch list options.
- Do not pass `wrapperSourceLink` and `wrapperSourcePath` with `acceptedValuesList` prop
- Create accepted values list for the item status field

## Issue
[UIDATIMP-515](https://issues.folio.org/browse/UIDATIMP-515)

## Screenshots
**Before fix**
image
**After fix**
image


src/settings/MappingProfiles/MappingProfilesForm.js
```diff
@@ -162,6 +162,22 @@ export const MappingProfilesFormComponent = ({
...
+        { name: 'In process' },
+        { name: 'In transit' },
+        { name: 'Missing' },
+        { name: 'On order' },
+        { name: 'Paged' },
+        { name: 'Declared lost' },
+        { name: 'Order closed' },
+        { name: 'Claimed returned' },
+      ],
+    },
```
IhorKoba, line 180: Maybe is it better to move these data somewhere to constants?

src/components/AcceptedValuesField/AcceptedValuesField.js
```diff
@@ -22,21 +24,24 @@ export const AcceptedValuesField = ({
   okapi,
   component,
   wrapperLabel,
+  acceptedValuesList,
```
VictorSoroka1, line 27:
```suggestion
  acceptedValuesList = [],
```

src/settings/MappingProfiles/MappingProfilesForm.js
```diff
@@ -162,6 +162,22 @@ export const MappingProfilesFormComponent = ({
       onLink: setAddedRelations,
       onUnlink: setDeletedRelations,
     },
+    'item-status': {
+      acceptedValuesList: [
+        { name: 'Available' },
```
VictorSoroka1, line 167: Since the values are shown on the UI it means they won't be localized. Does it make sense to create a key-value pair for having the keys (keys can be stored in constants) for coding and localization?
```
{
 key: 'AWAITING_PICKUP',
 value: <FormattedMessage ...
}
```

src/components/AcceptedValuesField/AcceptedValuesField.js
mariia-aloshyna: Added it to `defaultProps`

src/settings/MappingProfiles/MappingProfilesForm.js
mariia-aloshyna: Done

src/settings/MappingProfiles/MappingProfilesForm.js
mariia-aloshyna: Done

PR #548, opened 2020-05-20
UIDATIMP-497: Implement add new row functionality in MARCTable

mariia-aloshyna:
## Purpose
To create the Plus sign option that adds a row in the table of the Field Mapping profile details for MARC modifications.

## Approach
* Add `addNewRow` method in the `MARCTable` component.
* Insert new object into the rows array and change orders value for rows that are below in the table.
* Add `updateRowsData` method in the `MARCTable` component to handle all changes until it is connected to the redux form.
* Add tests

## Issue
[UIDATIMP-497](https://issues.folio.org/browse/UIDATIMP-497)

## Screenshot
add_row


src/components/MARCTable/MARCTableRow.js
```diff
@@ -286,8 +280,8 @@ export const MARCTableRow = ({
                 <FormattedMessage id="ui-data-import.settings.mappingProfile.marcTable.data.find" />
               </Label>
               <TextArea
-                value={dataValue?.find || ''}
-                onChange={e => onDataChange(e, 'find')}
+                value={field.data?.find || ''}
+                onChange={onDataFieldChange('find')}
```
VictorSoroka1, line 284: I think in the long run string literals params to this function are good to have in constants.

mariia-aloshyna: I hope when these fields will be wrapped in the Field and connected to the form, `onChange` methods will became unnecessary and literals as well. But if it is not so, then I'll move it to constants

VictorSoroka1: Sounds good.

PR #551, opened 2020-05-22
UIDATIMP-496: Implement re-order row functionality in the MARCTable

mariia-aloshyna:
## Purpose
To create the up/down arrows that resequence rows in the table of the Field Mapping profile details for MARC modifications

## Approach
* Add `moveRow` method to the `MARCTable` component
* Add test

## Ticket
[UIDATIMP-496](https://issues.folio.org/browse/UIDATIMP-496)

## Screenshot
reorder

### DIFF
--- a/src/components/MARCTable/MARCTable.js
+++ b/src/components/MARCTable/MARCTable.js
@@ -60,6 +60,25 @@ export const MARCTable = ({
     setRows(updatedRows);
   };
 
+  const moveRow = (order, orderToSwitch) => {
+    const rowIndex = rows.findIndex(field => field.order === order);
+    const rowToSwitchIndex = rows.findIndex(field => field.order === orderToSwitch);
+    const updatedRows = [...rows];
+
+    const rowToSwitch = {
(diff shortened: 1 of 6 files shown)


src/components/MARCTable/MARCTable.js
```diff
@@ -60,6 +60,28 @@ export const MARCTable = ({
     setRows(updatedRows);
   };
 
+  const moveRow = (order, orderToSwitch) => {
+    const rowIndex = rows.findIndex(field => field.order === order);
+    const rowToSwitchIndex = rows.findIndex(field => field.order === orderToSwitch);
+    const updatedRows = [...rows];
+
+    updatedRows[rowIndex] = {
```
VictorSoroka1, line 68: If you store to a variable this and the statement below
```
const row = updatedRows[rowIndex] = {...
const updatedRow = updatedRows[rowToSwitchIndex] = {...

updatedRows[rowIndex] = updatedRow;
updatedRows[rowToSwitchIndex] = row;
```
The name of variables I suggested is probably not the best though.

PR #552, opened 2020-05-25
UIDATIMP-500: Initial data doesn't show up for checkbox decorator for existing profiles

IhorKoba:
## Description
Initial data doesn't show up for checkbox decorator for existing profiles

## Approach
- Fix value of fields and paths in `initialData` for Mapping Profiles
- create `getBooleanActionsValue` function in utils for getting correct option value for `withBooleanActions` decorator
- update tests for `withBooleanActions` decorator in MCL component

## Issue
[UIDATIMP-500](https://issues.folio.org/browse/UIDATIMP-500)


src/components/FlexibleForm/Control.js
```diff
@@ -207,21 +206,11 @@ export const Control = memo(props => {
...
-          ...attrs,
-          value: newValue ? (<FormattedMessage id={newValue} />) : newValue,
-        };
-      }
-
       attrs = {
         ...attrs,
         component: wrapper || control,
         WrappedComponent: wrapper ? control : null,
+        value: getBooleanActionsValue(attrs?.value) || noValueComponent,
```
mariia-aloshyna, line 213: What happens if the decorator is any other, not the withBooleanActions?

src/components/FlexibleForm/ControlDecorators/withBooleanActions/withBooleanActions.js
```diff
@@ -35,12 +35,13 @@ const WithBooleanActionsControl = props => {
   } = props;
 
   const actions = get(FORMS_SETTINGS, [ENTITY_KEYS.MAPPING_PROFILES, 'DECORATORS', 'BOOLEAN_ACTIONS'], []);
-  const dataOptions = actions.map(action => ({
+  const dataOptions = actions.OPTIONS.map(action => ({
     value: action.value,
     label: intl.formatMessage({ id: action.label }),
   }));
+  const defaultValue = input?.value && dataOptions.filter(option => option.value === input.value).length ? input.value : '';
```
mariia-aloshyna, line 42: Move `input?.value && dataOptions.filter(option => option.value === input.value).length` condition to a separate variable, please

src/utils/getBooleanActionsValue.js
```diff
@@ -0,0 +1,30 @@
...
+import {
+  ENTITY_KEYS,
+  FORMS_SETTINGS,
+  isFormattedMessage,
+  isTranslationId,
+} from '.';
+
+export const getBooleanActionsValue = value => {
+  const actions = get(FORMS_SETTINGS, [ENTITY_KEYS.MAPPING_PROFILES, 'DECORATORS', 'BOOLEAN_ACTIONS'], []);
+  const newValue = actions.OPTIONS.find(item => item.value === value)?.label;
```
mariia-aloshyna, line 15:
```suggestion
  const actionLabel = actions.OPTIONS.find(item => item.value === value)?.label;
```

src/components/FlexibleForm/ControlDecorators/withBooleanActions/withBooleanActions.js
VictorSoroka1, line 42:
```suggestion
  const defaultValue = input?.value && dataOptions.some(option => option.value === input.value) ? input.value : '';
```

src/components/FlexibleForm/Control.js
IhorKoba: This function converts value only in case when it's Boolean Action Value, for others it returns value as is

mariia-aloshyna: In this case, maybe it will be better to rename `getBooleanActionsValue` method if it returns a value for any type of decorator not just for `withBooleanActions`?

IhorKoba: If you have a good name for that function, please advice.

src/utils/getDecoratorValue.js
```diff
@@ -0,0 +1,30 @@
...
+
+import {
+  ENTITY_KEYS,
+  FORMS_SETTINGS,
+  isFormattedMessage,
+  isTranslationId,
+} from '.';
+
+export const getDecoratorValue = value => {
+  const actions = get(FORMS_SETTINGS, [ENTITY_KEYS.MAPPING_PROFILES, 'DECORATORS', 'BOOLEAN_ACTIONS'], []);
```
mariia-aloshyna, line 14: As far as this method can process values for different types of decorators, consider renaming this const to `booleanActions`

PR #561, opened 2020-06-02
UIDATIMP-514: Add location code display to accepted values list

IhorKoba:
## Overview
In Holdings and Item field mapping profiles, reference dropdown lists for Permanent and Temporary locations should display the name of the location, followed by the location code in parentheses.

## Approach
* Create getAcceptedValuesOption function in utils to have a possibility to extend current accepted values option with additional data.
* Check incoming value, update it in appropriate way

## Ticket
[UIDATIMP-514](https://issues.folio.org/browse/UIDATIMP-514)

## Before
image

## After
image


src/components/AcceptedValuesField/AcceptedValuesField.js
```diff
@@ -39,7 +40,7 @@ export const AcceptedValuesField = ({
     data.forEach(item => {
       acceptedValues = {
         ...acceptedValues,
-        [item.id]: item.name,
+        [item.id]: getAcceptedValuesOption(item, 'name', 'code'),
```
mariia-aloshyna, line 43: Consider retrieving `optionField` and `fieldToAdd` values from the props because it will not be working if you will need to concatenate 'name' and some other field

src/components/FlexibleForm/ControlDecorators/partials/OptionsList.js
```diff
@@ -51,12 +55,12 @@ export const OptionsList = ({
                 role="menuitem"
                 align="start"
                 onClick={() => {
-                  onSelect(option?.[optionValue] || option?.[optionLabel]);
+                  onSelect(getAcceptedValuesOption(option, optionToSelect, 'code'));
```
mariia-aloshyna, line 58: Same here. Option list has not to know about 'code', it should be flexible. Maybe it makes sense to initialize options values outside and pass it to this component?

mariia-aloshyna:
Could you check tests, please, and update if needed?

src/components/AcceptedValuesField/AcceptedValuesField.js
IhorKoba: It's not a value, it's name of field, from which value should be taken

mariia-aloshyna: I see, I mean that field names can be different, e.g. for one field you need to concatenate 'name' and 'code', for another one you may need to concatenate 'name' and 'someFieldName'

mariia-aloshyna: Consider passing something like the template string and parse it in this component. For example:
`optionTemplate: '${name} (${code})'` 
We have something similar to it in the Control component, where we replace `**ri**` with `repeatableIndex`

src/utils/getAcceptedValuesOption.js
```diff
@@ -0,0 +1,25 @@
+const getValueFromTeplate = (option, template) => {
+  let newTemplate = template;
+
+  const pattern = /[**\w**]+/g;
+  const fields = newTemplate.match(pattern);
+
+  fields.forEach(field => {
+    const curentField = field.replace(/[**]+/g, '');
```
IvanKryzhanovskyi, line 8:
```suggestion
    const currentField = field.replace(/[**]+/g, '');
```

src/components/AcceptedValuesField/AcceptedValuesField.js
```diff
@@ -39,20 +41,26 @@ export const AcceptedValuesField = ({
...
         ...acceptedValues,
-        [item.id]: item.name,
+        [item.id]: optionTemplate ? updateValueWithTemplate(item, optionTemplate) : item[optionValue],
       };
     });
 
     return acceptedValues;
   };
 
+  const updateListOption = data => data.map(option => ({
```
mariia-aloshyna, line 51: Just suggestion
```suggestion
  const getListOptions= data => data.map(option => ({
```

IhorKoba: As for me we are updating options, that’s why I named it this way

PR #563, opened 2020-06-04
UIDATIMP-514: SyntaxError: invalid regexp group in FF :: Fixed

IhorKoba:
## Approach
* Get rid of lookbehind usage in regular expression

## Ticket
[UIDATIMP-514](https://issues.folio.org/browse/UIDATIMP-514)


src/utils/updateValueWithTemplate.js
```diff
@@ -1,15 +1,20 @@
...
 
   let updatedString = template;
   const match = template.match(findPattern);
 
   if (!isEmpty(match)) {
     match.forEach(key => {
-      updatedString = updatedString.replace(replacePattern, option[key]);
+      const currentKey = key.replace(/\*\*/g, '');
+      const replacePattern = new RegExp(`\\**${currentKey}\\**`, 'g');
+      const templateHasOption = !!template.match(currentKey);
```
mariia-aloshyna, line 13:
```suggestion
      const templateHasOptions = !isEmpty(template.match(currentKey));
```

src/utils/updateValueWithTemplate.js
```diff
@@ -1,15 +1,20 @@
...
+  const findPattern = /\*\*(\w+)\*\*/g;
 
   let updatedString = template;
   const match = template.match(findPattern);
 
   if (!isEmpty(match)) {
     match.forEach(key => {
-      updatedString = updatedString.replace(replacePattern, option[key]);
+      const currentKey = key.replace(/\*\*/g, '');
+      const replacePattern = new RegExp(`\\**${currentKey}\\**`, 'g');
```
mariia-aloshyna, line 12: Are you sure you need escape characters here?

IhorKoba: according to developer.mozilla.org, yes - 
image

mariia-aloshyna: In this case, maybe it's needed to escape second asterisk as well?
```suggestion
      const replacePattern = new RegExp(`\\*\\*${currentKey}\\*\\*`, 'g');
```

IhorKoba: I've tested and it works the same in both cases..

mariia-aloshyna: Did you try to match smth like this - `****code*****`?

IvanKryzhanovskyi: should we really cover such case?

IhorKoba: you're right, not the same. escaped second asterisk as well.

src/utils/updateValueWithTemplate.js
IhorKoba: What is the difference, except one more method is used ?

mariia-aloshyna: As for me, it looks a little bit more readable, but it's up to you actually

IhorKoba: ok

PR #566, opened 2020-06-09
UIDATIMP-536: Connect MARC modifications table to the form

mariia-aloshyna:
## Purpose
To edit and save the Field Mapping profile details for MARC modifications

## Approach

- Wrap form components into the `Field` component.
- Dispatch `marcMappingDetails` field change on `add`, `delete`, `reorder` row.
- Add tests.

## Ticket
[UIDATIMP-536](https://issues.folio.org/browse/UIDATIMP-536)


src/components/MARCTable/MARCTableRowContainer.js
```diff
@@ -13,38 +11,41 @@ export const MARCTableRowContainer = ({
...
           onRemoveRow={onRemoveRow}
           onMoveRow={onMoveRow}
-          onDataChange={onDataChange}
-          intl={intl}
+          subfieldIndex={0}
         />
         {containsSubsequentLines &&
-          field.subfields.map(subfield => (
+          data.field.subfields.map((subfield, idx) => idx !== 0 && (
             <div>
```
VictorSoroka1, line 40: Since this is a `map`, should we have some sort of `key` attribute in this case?

src/settings/MappingProfiles/initialDetails/index.js
```diff
@@ -31,16 +33,21 @@ export const getInitialDetails = (entity, stripRepeatableFields = false) => {
...
 
+  if (entityIsMARC) {
+    return currentEntity;
+  }
+
   const fields = currentEntity.mappingFields.map(field => ({
     ...field,
     subfields: [],
-  }));
+  })) || [];
```
VictorSoroka1, line 50: Is `|| []` really needed?

src/settings/MappingProfiles/initialDetails/index.js
```diff
@@ -31,16 +33,21 @@ export const getInitialDetails = (entity, stripRepeatableFields = false) => {
...
   }
 
+  const entityIsMARC = isMARCType(entity);
   const currentEntity = initialValues[entity];
 
   if (!stripRepeatableFields) {
     return currentEntity;
   }
 
+  if (entityIsMARC) {
```
VictorSoroka1, line 43: I think this can be combined with the condition above.

src/components/MARCTable/MARCTableRow.js
```diff
@@ -283,11 +258,11 @@ export const MARCTableRow = ({
               style={findContainerStyle}
             >
               <Label style={labelStyle}>
-                <FormattedMessage id="ui-data-import.settings.mappingProfile.marcTable.data.find" />
+                {formatMessage({ id: 'ui-data-import.settings.mappingProfile.marcTable.data.find' })}
```
VictorSoroka1, line 261: Could you please clarify why you started using this approach? I would say that `formatMessage` approach should be used only when `<FormattedMessage />` is not suited.

mariia-aloshyna: I used `formatMessage` for rendering such props as `placeholder`, option label and so on, so I decided just get rid of `FormattedMessage` import and use `formatMessage` instead in this component. There is no problem to return `FormattedMessage`. Do you think this approach is better?

VictorSoroka1: I remember there were a discussion about getting away from `formatMessage` approach. When I find a link I will provide it here. Only based on that I would recommend prefer using `FormattedMessage` whenever possible like in this case. Once it is not, then `formatMessage` is good to go.

mariia-aloshyna: I got it, thank you!

PR #592, opened 2020-06-22
Revert mapping holdings details view section

mariia-aloshyna:


src/settings/MappingProfiles/detailsSections/view/HoldingsDetailsSections/HoldingsNotes.js
```diff
@@ -0,0 +1,89 @@
...
+  const notesMapping = {
+    noteType: (
+      <FormattedMessage id={`${TRANSLATION_ID_PREFIX}.field.notes.noteType`} />
+    ),
+    note: (
+      <FormattedMessage id={`${TRANSLATION_ID_PREFIX}.field.notes.note`} />
+    ),
+    staffOnly: (
+      <FormattedMessage id={`${TRANSLATION_ID_PREFIX}.field.notes.staffOnly`} />
+    ),
```
Vladyslav-Velytskyi, line 40: Here is not so long strings to split them, you can write each value in save string with each key

mariia-aloshyna: Sorry, but it can't be written in one string due to eslint warnings

PR #593, opened 2020-06-22
UIDATIMP-429: Fix Accessibility problems for /data-import page

IhorKoba:
## Purpose
To fix all Accessibility problems for /data-import page

## Approach
- Change tag inside list from `<div>` to `<li>` in `Job` component
- Update styles

## Ticket
[UIDATIMP-429](https://issues.folio.org/browse/UIDATIMP-429)

### DIFF
--- a/src/components/Jobs/components/Job/Job.css
+++ b/src/components/Jobs/components/Job/Job.css
@@ -20,6 +20,15 @@
   }
 }
 
+[class*="listStyleDefault"] {
+  & .job {
+    margin-bottom: 0;
+    display: block;
+    padding: .5rem;
+    border-radius: var(--radius);
+  }
+}
+
 .jobHeader {
   font-size: var(--font-size-medium);
   font-weight: 600;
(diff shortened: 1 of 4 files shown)


src/components/Jobs/components/Job/Job.css
```diff
@@ -20,6 +20,15 @@
   }
 }
 
+[class*="listStyleDefault"] {
```
mariia-aloshyna, line 23: Will it work for `.listStyleDefault`?

IhorKoba: This styles apply only for `.job` inside `. listStyleDefault ` and it's only this case.

mariia-aloshyna: I mean will it work if `[class*="listStyleDefault"]` replace with `.listStyleDefault`?

IhorKoba: no, because of hash added to styles for parent component

mariia-aloshyna: Got it, thanks

PR #632, opened 2020-07-24
UIDATIMP-586: Introduce react-testing-library + Jest

mariia-aloshyna:
## Description
This PR introduces the use of `react-testing-library` + `jest` for unit testing.
`MARCTable` component was covered with unit tests.

## Existing problem
The coverage report is overwritten. The question is how to join e2e coverage report with unit tests coverage report.

## Result of passing tests

test_run

## Report by Jest about covering component with tests
MARCTable_coverage


Jenkinsfile
```diff
@@ -4,5 +4,5 @@ buildNPM {
   runLint = 'yes'
   runSonarqube = true
   runTest = 'yes'
-  runTestOptions = '--karma.singleRun --karma.browsers ChromeDocker --karma.reporters mocha junit --coverage'
+  runTestOptions = ''
```
VictorSoroka1, line 7: This can be removed at all, right?

test/jest/__mock__/reduxForm.mock.js
```diff
@@ -0,0 +1,38 @@
...
+import {
+  createStore,
+  combineReducers,
+} from 'redux';
+import { Provider } from 'react-redux';
+
+export const reduxFormMock = (component, initialStateValues = {}, formFieldValues = {}) => {
+  const onSubmit = jest.fn();
+  const fieldReducer = (state = initialStateValues, { type }) => {
+    switch (type) {
```
VictorSoroka1, line 15: `switch` can be omitted at all.

package.json
```diff
@@ -14,27 +14,37 @@
   "scripts": {
     "start": "stripes serve",
     "build": "stripes build --output ./output",
-    "test": "stripes test karma",
-    "test-coverage": "stripes test karma --coverage",
+    "test:unit": "jest --ci --coverage",
+    "test:e2e": "stripes test karma --karma.singleRun --karma.browsers ChromeDocker --karma.reporters mocha junit --coverage",
+    "test": "yarn run test:unit && yarn run test:e2e",
```
VictorSoroka1, line 19: I think with a new approach in order to have a proper coverage metric only the jest tests would be needed. As biggest also produces coverage information. And I think in this case two tools might collide. Overall I think as the result of this work PR should produce some coverage in the sonar report. Try to play around with just jest tests on CI.

src/components/MARCTable/MARCTableRow.test.js
```diff
@@ -0,0 +1,71 @@
...
+        onFieldUpdate={onFieldUpdate}
+        removeSubfieldRows={removeSubfieldRows}
+        removeSubactionFromRow={removeSubactionFromRow}
+        removeDataValuesFromRow={removeDataValuesFromRow}
+        fillEmptyFieldsWithValue={fillEmptyFieldsWithValue}
+      />
+    </Intl>
+  );
+
+  return render(reduxFormMock(component));
```
VictorSoroka1, line 42: I think in the future it makes sense to create a custom render component which at least wraps children in the Intl components as down the road it should then be added to every test.

src/components/MARCTable/MARCTable.test.js
```diff
@@ -0,0 +1,352 @@
...
+      />
+    </Intl>
+  );
+
+  return render(reduxFormMock(component));
+};
+
+describe('MARC modifications table', () => {
+  afterEach(() => {
+    cleanup();
```
VictorSoroka1, line 37: Are you sure that cleanup is really needed? I was thinking that it should work by default.

src/components/MARCTable/MARCTable.test.js
```diff
@@ -0,0 +1,352 @@
+import React from 'react';
+import {
+  render,
+  cleanup,
```
VictorSoroka1, line 4: The same point about cleanup.

https://testing-library.com/docs/react-testing-library/setup#skipping-auto-cleanup

src/components/MARCTable/MARCTable.test.js
```diff
@@ -0,0 +1,352 @@
...
+    describe('when subaction selected', () => {
+      it('there should be "Data" field', () => {
+        const { getByTestId } = renderMARCTable(fields);
+
+        fireEvent.change(getByTestId('marc-table-subaction'), { target: { value: 'CREATE_NEW_FIELD' } });
+
+        expect(getByTestId('marc-table-data')).toBeDefined();
+      });
+
+      it('there should be "Data" field', () => {
```
VictorSoroka1, line 343: Please make the message more descriptive as the test above it has the same.

package.json
mariia-aloshyna: This is the thing I am thinking about. For now, the coverage report can be generated only for e2e tests or unit tests because of different reporters. However, I noticed that ui-receiving module has such kind of [script](https://github.com/folio-org/ui-receiving/blob/master/package.json#L160), so I decided to try it for ui-data-import. But yeah, report's still overwritten.

VictorSoroka1: I see that ui-receiving run unit tests after BigTest. So maybe this is also a thing to try out.

mariia-aloshyna: Already tried. When running unit tests after BigTest then there is information only about unit tests coverage (MARCTable component). When running them before then there is information only about e2e tests coverage (MARCTable component coverage is about 0%). One report overwrites another.

mariia-aloshyna: So I guess that it happens because of different reporters, but I still haven't figured out how to use the same reporter for both tests.

VictorSoroka1: I think it is great that jest produces information. I think BigTests is enough to just run really without bothering about coverage aspect. Because of the unstable tests, the coverage for it would always be zero if at least one test fails.

src/components/MARCTable/MARCTableRow.test.js
mariia-aloshyna: I'm not sure I understand what you mean, could you clarify, please?

VictorSoroka1: I mean creating some sort of test setup component which wraps the content in intlProvider so in future it won't be duplicated over and over.

```
const renderWithIntl = { children } => {
   return render(<Intl intlProps={<some intl props if needed>}>{props.childrend})
}
```

mariia-aloshyna: Done

package.json
mariia-aloshyna: @VictorSoroka1 FYI Ian created a ticket for investigating this- https://issues.folio.org/browse/FOLIO-2703

PR #639, opened 2020-07-30
UIDATIMP-535: Field Mapping Profile details: Have view details screen cover the entire screen

IhorKoba:
## Purpose
To provide a full screen view of the field mapping profile View details

## Approach
- Use  `FullScreenView` component from stripes-data-transfer-components
- add detecting `isMappingProfile` to render Mapping profile details in another way
- update styles

## Issue
[UIDATIMP-535](https://issues.folio.org/browse/UIDATIMP-535)

### DIFF
--- a/src/components/SearchAndSort/SearchAndSort.js
+++ b/src/components/SearchAndSort/SearchAndSort.js
@@ -138,6 +138,7 @@ export class SearchAndSort extends Component {
     finishedResourceName: PropTypes.string,
     fullWidthContainer: PropTypes.instanceOf(Element),
     rowUpdater: PropTypes.func,
+    isFullScreen: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -462,8 +463,30 @@ export class SearchAndSort extends Component {
       fullWidthContainer,
       handleEditSuccess,
       ViewRecordComponent,
+      isFullScreen,
       location: { pathname },
     } = this.props;
+
+    const commonProps = {
+      key: pathname,
+      stripes,
+      editContainer: fullWidthContainer,
+      parentResources,
+      connectedSource: source,
+      parentMutato: parentMutator,
+      onClose: this.collapseRecordDetails,
+      onCloseEdit: this.onCloseEditRecord,
+      onEdit: this.editRecord,
+      onDelete: this.deleteRecord,
+      onEditSuccess: handleEditSuccess,
+    };
+
+    const narrowViewScreenProps = {
+      ...commonProps,
+      paneWidth: '44%',
+    };
+
+    const viewRecordComponentProps = isFullScreen ? commonProps : narrowViewScreenProps;
 
     return (
       <Route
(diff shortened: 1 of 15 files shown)


src/settings/MappingProfiles/ViewMappingProfile.js
```diff
@@ -219,11 +227,10 @@ export class ViewMappingProfile extends Component {
...
 
     return (
-      <Pane
-        id={paneId}
-        defaultWidth="fill"
-        fluidContentWidth
+      <FullScreenView
+        paneId={paneId}
         renderHeader={this.renderPaneHeader}
+        className={sharedCss.fullScreenDetails}
```
VictorSoroka1, line 233: Please try the approach that ui-data-export uses in order to get rid of this kind of style. I think the usage of Layer can avoid such CSS files. Up to you really.

https://github.com/folio-org/ui-data-export/blob/master/src/settings/ProfileDetails/ProfileDetails.js#L60

IhorKoba: Thanks! Using Layer helps to avoid additional styles.

src/components/SearchAndSort/SearchAndSort.js
```diff
@@ -725,28 +726,34 @@ export class SearchAndSort extends Component {
   };
 
   render() {
+    const { finishedResourceName } = this.props;
+
+    const isMappingProfile = finishedResourceName === ENTITY_KEYS.MAPPING_PROFILES;
```
mariia-aloshyna, line 731: Maybe it makes sense to not get tied to the `Mapping profiles` entity and pass some prop like 'isDetailsPaneFullScreen' from the root (`<MappingProfiles>`)? What do you think?

src/components/SearchAndSort/SearchAndSort.js
```diff
@@ -725,28 +726,34 @@ export class SearchAndSort extends Component {
...
-          </div>
-        </Pane>
-        {this.renderDetailsPane(source)}
-        {this.renderCreateRecordLayer(source)}
-      </Paneset>
+          </Pane>
+          {!isMappingProfile && this.renderDetailsPane(source)}
+          {this.renderCreateRecordLayer(source)}
+        </Paneset>
+        {isMappingProfile && this.renderDetailsPane(source)}
```
mariia-aloshyna, line 755: Look [there](https://github.com/folio-org/ui-data-import/pull/639/files#diff-7e8ccabec46feaba584bac318c32f358R475-R486) are props which `<FullScreenView>` component doesn't need, so consider splitting props for the full screen and for another one.

src/components/SearchAndSort/SearchAndSort.js
```diff
@@ -462,27 +462,38 @@ export class SearchAndSort extends Component {
...
+      connectedSource: source,
+      parentMutato: parentMutator,
+      onClose: this.collapseRecordDetails,
+      onCloseEdit: this.onCloseEditRecord,
+      onEdit: this.editRecord,
+      onDelete: this.deleteRecord,
+      onEditSuccess: handleEditSuccess,
+    };
+
+    const narrowScreenViewProps = {
```
mariia-aloshyna, line 483:
```suggestion
    const narrowViewScreenProps = {
```

src/settings/MappingProfiles/MappingProfiles.js
```diff
@@ -220,6 +221,7 @@ export class MappingProfiles extends Component {
       description: '',
       mappingDetails: {},
     },
+    isFullScreen: true,
```
mariia-aloshyna, line 224:
```suggestion
    isViewFullScreen: true,
```

PR #667, opened 2020-08-31
UIDATIMP-613: MARC Bib field mapping profile: details for Update Selected fields on Create/Edit screen

IhorKoba:
## Purpose
To indicate that only certain fields in an SRS MARC record should be updated, instead of replacing the entire record, when a user is importing

## Approach
* Update logic of rendering cells in row to depend on passed headings
* Add `renderMARCTable` method to `MappingProfilesForm` component
* Add `ConfirmationModal ` to `MappingProfilesForm` to confirm change action of MARC type select
* Move `fillEmptyFieldsWithValue` to external `util` to be able to re-use it in different places
* Update unit tests

## Ticket
[UIDATIMP-613](https://issues.folio.org/browse/UIDATIMP-613)

## Screenshot
sdIYOAuvxB


src/components/MARCTable/MARCTable.js
```diff
@@ -212,10 +198,11 @@ export const MARCTable = ({
       className={css.tableContainer}
     >
       <MARCTableHeader
-        columns={columns}
+        columns={visibleColumns || columns}
```
ViktorSoroka07, line 201: I would suggest moving `columns` to defaultProps and deal only with one variable/prop for that.
```suggestion
        columns={columns}
```

ViktorSoroka07 (approved):

IvanKryzhanovskyi (approved):

PR #671, opened 2020-09-03
UIDATIMP-645: Validation function validateRequiredFields works incorrectly after latest update

IhorKoba:
## Purpose
To fix `validateRequiredFields` function

## Approach
- Update validation for subfield field for `MARCFieldProtection` component
- revert changes for common `validateRequiredFields` function

## Tickets
[UIDATIMP-645](https://issues.folio.org/browse/UIDATIMP-645)

### DIFF
--- a/src/settings/MARCFieldProtection/MARCFieldProtection.js
+++ b/src/settings/MARCFieldProtection/MARCFieldProtection.js
@@ -8,10 +8,7 @@ import {
 import { stripesConnect } from '@folio/stripes/core';
 import { ControlledVocab } from '@folio/stripes/smart-components';
 
-import {
-  MARC_FIELD_PROTECTION_SOURCE,
-  validateRequiredField,
-} from '../../utils';
+import { MARC_FIELD_PROTECTION_SOURCE } from '../../utils';
 
 const validateField = value => {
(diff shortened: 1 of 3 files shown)


IvanKryzhanovskyi (approved):

src/settings/MARCFieldProtection/MARCFieldProtection.js
```diff
@@ -39,12 +36,16 @@ const validateSubfield = (subfieldValue, fieldValue) => {
   let pattern = /^[*\w]$/;
   let errorMessage = <FormattedMessage id="ui-data-import.validation.enterAsteriskOrAlphanumeric" />;
 
-  if (fieldValue === '*') {
+  if (!subfieldValue) {
+    errorMessage = <FormattedMessage id="stripes-core.label.missingRequiredField" />;
+  }
+
+  if (subfieldValue && fieldValue === '*') {
```
ViktorSoroka07, line 43:
```suggestion
  else if (fieldValue === '*') {
```
I think here the code can be simplified as in case the `subfieldValue` is falsy the condition above will take place.

ViktorSoroka07 (approved):

PR #675, opened 2020-09-09
UIDATIMP-614: MARC Bib field mapping profile: details for Update Selected fields on View screen

IhorKoba:
## Purpose
To indicate that only certain fields in an SRS MARC record should be updated, instead of replacing the entire record, when a user is importing

## Approach
* Update logic of rendering cells in row to depend on passed headings
* Add `renderMARCTableView` method to `ViewMappingProfile` component
* Add `getCellValue` method to `MARCTableViewRow` component

## Ticket
[UIDATIMP-614](https://issues.folio.org/browse/UIDATIMP-614)

## Screenshot

image

### DIFF
--- a/src/components/MARCTableView/MARCTableViewRow.js
+++ b/src/components/MARCTableView/MARCTableViewRow.js
@@ -4,6 +4,8 @@ import { FormattedMessage } from 'react-intl';
 
 import { isEmpty } from 'lodash';
 
+import { NoValue } from '@folio/stripes/components';
+
 import {
   MAPPING_DETAILS_ACTIONS,
   MAPPING_DETAILS_SUBACTIONS,
@@ -11,6 +13,7 @@ import {
   SUBACTION_OPTIONS,
   POSITION_OPTIONS,
   mappingMARCViewFieldShape,
+  getTrimmedValue,
 } from '../../utils';
 
 import css from '../MARCTable/MARCTable.css';
@@ -36,8 +39,14 @@ export const MARCTableViewRow = ({
     subaction,
   } = rowData;
 
+  const getCellValue = value => {
+    const formattedValue = getTrimmedValue(value);
+
+    return !isEmpty(formattedValue) ? formattedValue : <NoValue />;
+  };
+
   const renderCell = (key, i) => {
-    const cellData = rowData[key];
+    const cellData = getCellValue(rowData[key]);
     const cellStyle = { width: columnWidths[key] };
 
     const getCellContent = () => {
@@ -158,7 +167,7 @@ export const MARCTableViewRow = ({
         className={css.tableCell}
         style={cellStyle}
       >
-        {!isEmpty(cellData) && getCellContent()}
+        {getCellContent()}
       </div>
     );
   };
--- a/src/components/MARCTableView/MARCTableViewRowContainer.js
+++ b/src/components/MARCTableView/MARCTableViewRowContainer.js
@@ -8,6 +8,7 @@ import { mappingMARCFieldShape } from '../../utils';
 import css from '../MARCTable/MARCTable.css';
 
 export const MARCTableViewRowContainer = ({
+  columns,
   fields,
   columnWidths,
 }) => {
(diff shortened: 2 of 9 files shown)


src/components/MARCTableView/MARCTableViewRow.js
```diff
@@ -36,8 +38,14 @@ export const MARCTableViewRow = ({
     subaction,
   } = rowData;
 
+  const getCellValue = value => {
+    const formattedValue = value && value.trim ? value.trim() : value;
```
mariia-aloshyna, line 42: I see this code is repeated in several places. Consider moving this to `utils`.

src/components/MARCTableView/MARCTableViewRowContainer.js
```diff
@@ -58,6 +66,7 @@ export const MARCTableViewRowContainer = ({
 };
 
 MARCTableViewRowContainer.propTypes = {
+  columns: PropTypes.arrayOf(PropTypes.string),
```
mariia-aloshyna, line 69: Looks like this prop is required

src/settings/MappingProfiles/ViewMappingProfile.js
```diff
@@ -214,11 +217,29 @@ export class ViewMappingProfile extends Component {
 
     const isMARCRecord = existingRecordType === MARC_TYPES.MARC_BIBLIOGRAPHIC;
 
+    const updatesFieldMappingForMARCColumns = ['field', 'indicator1', 'indicator2', 'subfield'];
+
+    const rebderMARCTableView = () => {
+      switch (marcMappingOption) {
```
mariia-aloshyna, line 223: Are you sure that `switch` is needed here?

src/components/MARCTableView/MARCTableViewRow.js
IhorKoba: agree!

src/components/MARCTableView/MARCTableViewRowContainer.js
IhorKoba: done

src/settings/MappingProfiles/ViewMappingProfile.js
IhorKoba: As for me, it's a little more readable and easy to extend with one option in case if needed. But If you think if/else is better here I don't mind to change it.

mariia-aloshyna: I'd suggest creating a variable for the default set of columns
```
const defaultFieldMappingForMARCColumns = ['action', 'field', 'indicator1', 'indicator2',
    'subfield', 'subaction', 'data', 'position']
```
And then use it smth like this
```
const renderMARCTableView = () => {
  const fieldMappingForMARCColumns = {
    UPDATES: updatesFieldMappingForMARCColumns,
    ...
  };

  return (
    <MARCTableView
          columns={fieldMappingForMARCColumns[marcMappingOption] || defaultFieldMappingForMARCColumns}
          fields={mappingDetails?.marcMappingDetails}
     />
  )
}; 
```

Don't know if this solution is better than `if/else` - it's up to you, but I guess the `switch` operator is not quite suitable for 1 or 2 options, what do you think?

src/settings/MappingProfiles/ViewMappingProfile.js
```diff
@@ -214,11 +217,29 @@ export class ViewMappingProfile extends Component {
 
     const isMARCRecord = existingRecordType === MARC_TYPES.MARC_BIBLIOGRAPHIC;
 
+    const updatesFieldMappingForMARCColumns = ['field', 'indicator1', 'indicator2', 'subfield'];
+
+    const rebderMARCTableView = () => {
```
mariia-aloshyna, line 222:
```suggestion
    const renderMARCTableView = () => {
```

src/settings/MappingProfiles/ViewMappingProfile.js
```diff
@@ -214,11 +217,26 @@ export class ViewMappingProfile extends Component {
 
     const isMARCRecord = existingRecordType === MARC_TYPES.MARC_BIBLIOGRAPHIC;
 
+    const defaultFieldMappingForMARCColumns = ['action', 'field', 'indicator1', 'indicator2',
+      'subfield', 'subaction', 'data', 'position'];
+    const updatesFieldMappingForMARCColumns = ['field', 'indicator1', 'indicator2', 'subfield'];
```
mariia-aloshyna, line 222: As far as those constants are used only in `renderMARCTableView()` consider moving them into the method.

mariia-aloshyna (approved):

src/settings/MappingProfiles/ViewMappingProfile.js
IhorKoba: done

IvanKryzhanovskyi (approved):

PR #694, opened 2020-09-29
UIDATIMP-630: Refine an identifier matching for Instances

mariia-aloshyna:
## Description
Data import users need to be able to match on a wide variety of identifiers in the Instance record. The current matching setup (Identifier and Identifier type as 2 separate matches) is cumbersome and may not work so it's needed to be able to match an identifying number from any MARC field to a particular identifier type number in an Instance (e.g. only match against OCLC number).

## Approach
- Add a new request to the manifest in `MatchProfiles`
- Add `fromResources` field to an item in `fieldsConfig` that contains info about fields which should be retrieved from the `resources`
- Rework `getDropdownOptions` function so options can be built from the `resources`

## Ticket
[UIDATIMP-630](https://issues.folio.org/browse/UIDATIMP-630)

## Screenshot
identifiers_matching


src/utils/getFieldsMatched.js
```diff
@@ -1,19 +1,77 @@
...
+
+  return records.find(record => record[fieldToSend] === fieldValue)?.[fieldToDisplay];
+};
+
+const getField = (fields, recordType, resources, intl) => {
+  if (!fields.length || !recordType) {
     return undefined;
   }
 
+  const fieldValue = fields.find(field => field.label === 'field')?.value;
```
Vladyslav-Velytskyi, line 25: Better to create constant for `'field'` string

src/components/ListView/ListView.js
```diff
@@ -247,6 +247,7 @@ export class ListView extends Component {
                 searchTerm,
                 selectRecord,
                 selectedRecords,
+                resources,
```
Vladyslav-Velytskyi, line 250: Do you use this value?

src/components/MatchCriterion/edit/ExistingRecordSections/ExistingSectionFolio.js
```diff
@@ -11,13 +11,43 @@ import {
...
+  const [isDirty, setDirty] = useState(false);
+
+  const dropdownValue = last(existingRecordFieldsValue).value;
+
+  const handleExistingRecordSelect = value => {
+    const fieldToChangeName = `profile.matchDetails[${repeatableIndex}].existingMatchExpression.fields`;
+    const fieldId = existingRecordFields.find(item => item.value === value)?.id;
+    const fieldFromConfig = fieldsConfig.find(item => item.id === fieldId);
+    const fieldToChangeValue = [{
+      label: 'field',
```
Vladyslav-Velytskyi, line 34: Const for `'field'`

src/components/ListView/ListView.js
mariia-aloshyna: Yep, here it is - https://github.com/folio-org/ui-data-import/pull/694/files#diff-eeed0328f61892b3be6ff4abb5f42dc1R71

Vladyslav-Velytskyi: I see, my bad

Vladyslav-Velytskyi (approved):
Good job!

src/components/MatchCriterion/edit/ExistingRecordSections/ExistingSectionFolio.js
```diff
@@ -44,11 +78,13 @@ export const ExistingSectionFolio = ({
 
 ExistingSectionFolio.propTypes = {
   repeatableIndex: PropTypes.number.isRequired,
+  existingRecordFieldsValue: PropTypes.arrayOf(PropTypes.object).isRequired,
   existingRecordFields: PropTypes.arrayOf(PropTypes.shape({
     value: PropTypes.string.isRequired,
     label: PropTypes.string.isRequired,
   })),
   existingRecordFieldLabel: PropTypes.node,
+  dispatchFormChange: PropTypes.func,
```
IhorKoba, line 87: Please, add DefaultProps for dispatchFormChange

IhorKoba (approved):

PR #696, opened 2020-10-02
UIDATIMP-651: Add capability to remove jobs that are stuck in "Running" area of Data Import landing page first pane

IhorKoba:
## Purpose
Sometimes jobs get stuck in the Running area of the Data Import landing page, when modules crash. They never finish, and they never disappear. We need a way to remove them from that area of the landing page. This story reuses most of the functionality we use when deleting a file that was not supposed to be uploaded.

## Approach
- Add possibility to send delete request to `Job` component
- Add styles for `Job` component in deleting state

## Ticket
[UIDATIMP-651](https://issues.folio.org/browse/UIDATIMP-651)

## Screenshot
image


src/components/Jobs/components/Job/Job.js
```diff
@@ -42,12 +75,74 @@ export class Job extends Component {
     return <span>{datePart} {timePart} {todayPart}</span>;
   }
 
+  createJobUrl = id => {
+    const { stripes: { okapi } } = this.props;
```
mariia-aloshyna, line 79:
```suggestion
    const { stripes: { okapi: { url: host } } } = this.props;
```

src/components/Jobs/components/Job/Job.js
```diff
@@ -42,12 +75,74 @@ export class Job extends Component {
...
+    const { url: host } = okapi;
+
+    return createUrl(`${host}/change-manager/jobExecutions/${id}/records`);
+  };
+
+  createDeleteJobHeaders = () => {
+    const { stripes: { okapi } } = this.props;
+
+    return createOkapiHeaders(okapi);
+  };
```
mariia-aloshyna, line 90: Are you sure you need this method? All it does is extract a prop and return another function, moreover, it's used only in `deleteJob` method.

src/utils/constants.js
```diff
@@ -5,6 +5,8 @@ export const STATUS_ERROR = 'ERROR';
 export const DEFAULT_FETCHER_UPDATE_INTERVAL = 5000;
 export const DEFAULT_TIMEOUT_BEFORE_FILE_DELETION = 0;
 
+export const DEFAULT_TIMEOUT_BEFORE_JOB_DELETION = 10000;
```
mariia-aloshyna, line 8: Can this const be also used in `src/routes/JobProfile` file?

src/components/Jobs/components/Job/Job.js
IhorKoba: agree,  removed

src/utils/constants.js
IhorKoba: done

mariia-aloshyna (approved):

PR #699, opened 2020-10-07
UIDATIMP-676: Match profile create-edit & view screens: change unusable options to disabled

IhorKoba:
## Purpose
To provide a visual cue for options that are not yet useable in the Job profile

## Approach
- Add array to `constants` with disabled Folio record types
- Add styles and functionality of disabling options in `RecordItem` component

## Ticket
[UIDATIMP-676](https://issues.folio.org/browse/UIDATIMP-676)

## Screenshot
image
image

### DIFF
--- a/src/components/RecordTypesSelect/components/RecordItem.js
+++ b/src/components/RecordTypesSelect/components/RecordItem.js
@@ -12,6 +12,8 @@ import classNames from 'classnames';
 import { Dropdown } from '@folio/stripes/components';
 import { AppIcon } from '@folio/stripes/core';
 
+import { FOLIO_RECORD_TYPES_TO_DISABLE } from '../../../utils';
+
 import { IncomingRecordMenu } from './IncomingRecordMenu';
 import { IncomingRecordTrigger } from './IncomingRecordTrigger';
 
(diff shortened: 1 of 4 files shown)


src/components/RecordTypesSelect/components/RecordItem.js
```diff
@@ -56,13 +58,21 @@ export const RecordItem = memo(({
...
       data-test-record-item
       tabIndex="0"
       role="button"
       id={incomingRecord.type}
-      className={classNames(css.item, { [css.clickableItem]: isEditable }, className)}
+      className={classNames(
+        css.item,
+        { [css.clickableItem]: isEditable },
+        className,
+        isOptionDisabled && css.disabledItem,
```
mariia-aloshyna, line 74: Consider adding a `disabled` field to `itemMeta` in the `<RecordSelect>` `recordsData` const instead of creating a `FOLIO_RECORD_TYPES_TO_DISABLE` const because `recordsData` is a config, so I think it makes sense to store such property there.

IhorKoba: As for me it's more convenient and logical to save all disabled options in one place, and use them in all profiles, where these options should be selected.

mariia-aloshyna (approved):

PR #711, opened 2020-10-11
only import MatchingFieldsManager once

zburke:
`MatchingFieldsManager` was imported twice.


zburke:
@IvanKryzhanovskyi, this error [broke the reference environments](https://folio-project.slack.com/archives/C210UCHQ9/p1602355113099600). I filed [UIDATIMP-690](https://issues.folio.org/browse/UIDATIMP-690) as a follow-up story since I would have expected lint to catch this.

PR #731, opened 2020-10-20
UIDATIMP-452: Fix Accessibility problems for settings/data-import/match-profiles (lists must only directly contain li elements)

IhorKoba:
## Purpose

To fix Accessibility problems for settings/data-import/match-profiles (lists must only directly contain li elements)

## Approach

- Change attribute `id` to `data-id` to avoid duplicating id's
- Change structure according to fixes

## Refs

[UIDATIMP-452](https://issues.folio.org/browse/UIDATIMP-452)

### DIFF
--- a/src/components/RecordTypesSelect/components/CompareRecordSelect.js
+++ b/src/components/RecordTypesSelect/components/CompareRecordSelect.js
@@ -59,7 +59,7 @@ export const CompareRecordSelect = memo(({
   }, []);
 
   useEffect(() => {
-    const selectedExistingRecordElem = getElement(`#${existingRecord.type}`, existingRecordsElemRef.current);
+    const selectedExistingRecordElem = getElement(`[data-id=${existingRecord.type}]`, existingRecordsElemRef.current);
 
     if (selectedExistingRecordElem) {
       setTop(selectedExistingRecordElem.offsetTop);
--- a/src/components/RecordTypesSelect/components/RecordItem.js
+++ b/src/components/RecordTypesSelect/components/RecordItem.js
@@ -64,9 +64,9 @@ export const RecordItem = memo(({
   const initialButton = (
     <div // eslint-disable-line jsx-a11y/click-events-have-key-events
       data-test-record-item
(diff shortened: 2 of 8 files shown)


src/components/RecordTypesSelect/components/CompareRecordSelect.js
```diff
@@ -59,7 +59,7 @@ export const CompareRecordSelect = memo(({
   }, []);
 
   useEffect(() => {
-    const selectedExistingRecordElem = getElement(`#${existingRecord.type}`, existingRecordsElemRef.current);
+    const selectedExistingRecordElem = getElement(`[data-id=${existingRecord.type}]`, existingRecordsElemRef.current);
```
mariia-aloshyna, line 62: Let's rename it to `data-testid` to be able to test it with `RTL` (see doc - https://testing-library.com/docs/dom-testing-library/api-queries#bytestid)

src/components/RecordTypesSelect/components/RecordItem.js
```diff
@@ -66,7 +66,7 @@ export const RecordItem = memo(({
       data-test-record-item
       tabIndex="0"
       role="button"
-      id={incomingRecord.type}
+      data-id={incomingRecord.type}
```
mariia-aloshyna, line 69: Move it to the top (to other `data` attributes), please

mariia-aloshyna (approved):

src/components/RecordTypesSelect/components/RecordItem.js
IhorKoba: done

src/components/RecordTypesSelect/components/CompareRecordSelect.js
IhorKoba: This data-id attribute uses for finding DOM element, and not for testing component. So, I suggest not to change this one to `data-testid` to avoid misunderstanding.

IvanKryzhanovskyi (approved):

PR #760, opened 2020-11-11
UIDATIMP-754 'Reset all' functionality leads to error page. Fixed

IvanKryzhanovskyi:
## Overview
'Reset all' functionality leads to error page

## Learning
First name is an optional field for user form. 
Sometimes jobProfile can be received as undefined

## Refs
https://issues.folio.org/browse/UIDATIMP-754


src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -159,11 +159,13 @@ class ViewAllLogs extends Component {
...
       .sort((userA, userB) => {
-        if (userA.firstName.localeCompare(userB.firstName) === 0) {
+        const name = userA.firstName || userA.lastName;
+
+        if (userA.firstName?.localeCompare(userB.firstName) === 0) {
           return userA.lastName.localeCompare(userB.lastName);
         }
 
-        return userA.firstName.localeCompare(userB.firstName);
+        return name.localeCompare(userB.firstName);
```
mariia-aloshyna, line 168: What happens when `userA` has `firstName` but `userB.firstName` === `undefined`?

IvanKryzhanovskyi: Then it should compare value with undefined and sort it, does it seems wrong behavior? Do you propose to have nameA and nameB instead to be sure that we cath all cases?

IvanKryzhanovskyi: Done

IhorKoba (approved):

mariia-aloshyna (approved):

PR #769, opened 2020-11-18
UIDATIMP-768: Field mapping for repeatable fields needs a tweak to validation

IhorKoba:
## Overview

For repeatable fields in the Instance, Holdings, and Item field mapping profiles, if you select an action first, FOLIO validates whether you add a value, or (if the action is remove all) does not allow you to add a value. If you add the value first (before selecting an action), the field mapping profile can be saved without an action for the repeatable field. And then the import errors because FOLIO does not know what action to take for the repeatable field.

## Approach

-Add validation rule for `RepeatableActionsField` component

## Refs

https://issues.folio.org/browse/UIDATIMP-768

### DIFF
--- a/translations/ui-data-import/en.json
+++ b/translations/ui-data-import/en.json
@@ -236,6 +236,7 @@
   "validation.quotationError": "Non-MARC value must use quotation marks",
   "validation.cannotBeUpdated": "This field cannot be updated",
   "validation.chooseDifferentField": "Please choose a different field",
+  "validation.selectAnAction": "Action must be selected before the profile can be saved",
   "validation.chooseAtLeastOneValue": "One or more values must be added before the profile can be saved.",
   "validation.enterAsteriskOrNumeric": "Please enter * or numeric value",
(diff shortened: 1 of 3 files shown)


IvanKryzhanovskyi (approved):

translations/ui-data-import/en.json
```diff
@@ -236,6 +236,7 @@
   "validation.quotationError": "Non-MARC value must use quotation marks",
   "validation.cannotBeUpdated": "This field cannot be updated",
   "validation.chooseDifferentField": "Please choose a different field",
+  "validation.selectAnAction": "Action must be selected before the profile can be saved",
```
mariia-aloshyna, line 239: Are you sure the validation message should look like this? In the ticket, I see "Select an action" warning.

IhorKoba: It's ok, discussed with Ann-Marie, and updated the description of the ticket in JIRA.

mariia-aloshyna (approved):

PR #771, opened 2020-11-23
UIDATIMP-755: Add "Load more" button to View all log page

IvanKryzhanovskyi:
## Purpose
To have a better UI for loading log entries
As a staff person working with data import logs
I want to be able to scroll through all the logs on the View all log page
So that I can find one that I want even if I don't know the exact name or date

## Approach
- property INITIAL_RESULT_COUNT has been substituted with RESULT_COUNT_INCREMENT for ViewAllLogs component;
- properties pagingType and pageAmount has been added to ViewAllLogs component configuration

## Refs
https://issues.folio.org/browse/UIDATIMP-755

## Screenshots
Load_more

### DIFF
--- a/src/routes/ViewAllLogs/ViewAllLogs.js
+++ b/src/routes/ViewAllLogs/ViewAllLogs.js
@@ -58,6 +58,7 @@ const visibleColumns = [
   'hrId',
 ];
 
+const INITIAL_RESULT_COUNT = 100;
 const RESULT_COUNT_INCREMENT = 100;
 
 @stripesConnect
@@ -84,7 +85,7 @@ class ViewAllLogs extends Component {
         sort: '-completedDate',
       },
     },
-    resultCount: { initialValue: RESULT_COUNT_INCREMENT },
+    resultCount: { initialValue: INITIAL_RESULT_COUNT },
     records: {
       type: 'okapi',
       clear: true,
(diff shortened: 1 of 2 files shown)


src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -216,7 +215,7 @@ class ViewAllLogs extends Component {
           packageInfo={this.props.packageInfo || packageInfo}
           objectName="job-logs"
           baseRoute={packageInfo.stripes.route}
-          initialResultCount={INITIAL_RESULT_COUNT}
+          initialResultCount={RESULT_COUNT_INCREMENT}
```
mariia-aloshyna, line 218: Are you sure we don't need this prop? This is a required one in `SearchAndSort`

src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -58,8 +58,7 @@ const visibleColumns = [
   'hrId',
 ];
 
-const INITIAL_RESULT_COUNT = 25;
```
mariia-aloshyna, line 61: Keep this const, please, even if it has the same value as `RESULT_COUNT_INCREMENT`, it has a different meaning. If you search through the project, you will see that this approach is used in many places.

IvanKryzhanovskyi: Done

mariia-aloshyna (approved):

PR #793, opened 2020-12-29
UIDATIMP-742 File upload in progress displays when it shouldn't. Fixed

IvanKryzhanovskyi:
## Overview
When uploading a file for import, after selecting a job profile, sometimes the users sees a "File upload in progress" screen, which should not display unless that user or another user is uploading a file on the same logon at the same time. It doesn't happen all the time, but pretty consistently.

## Approach
delay for server call has been added

## Refs
https://issues.folio.org/browse/UIDATIMP-742


src/components/ImportJobs/ImportJobs.js
```diff
@@ -62,7 +62,7 @@ export class ImportJobs extends Component {
   };
 
   componentDidMount() {
-    this.fetchUploadDefinition();
+    setTimeout(() => { this.fetchUploadDefinition(); }, 1000);
```
mariia-aloshyna, line 65: Cancel the timer in `componentWillUnmount()`, please.

IvanKryzhanovskyi: Done

mariia-aloshyna (approved):

PR #836, opened 2021-02-25
UIDATIMP-818 Create/Edit Match Profile WCAG 2.1 AA - Color Contrast violation

IvanKryzhanovskyi:
## Purpose
To fix warnings receiving from accessibility checker

## Approach
- Accordion header text color has been changed to #FFF
- Opacity for disabled items has been updated to 0.6 and the text color has been changed to #000.

## Refs
https://issues.folio.org/browse/UIDATIMP-818

## Screenshots
before
after


src/components/MatchCriterion/CommonMatchCriterion.css
```diff
@@ -5,6 +5,16 @@
     margin-top: 0;
   }
 
+  & [class|="headerInner"] {
+    & > span {
+        color: #b2b2b2;
+    }
+
+    & > div {
+      color: #ffffff;
```
mariia-aloshyna, line 14: Are there any problems with `var(--bg)` so you prefer to use HEX instead of it?
```suggestion
      color: var(--bg);
```

IhorKoba (approved):

mariia-aloshyna (approved):

PR #919, opened 2021-06-04
UIDATIMP-933 Update version of interfaces due to supporting MARC Auth…

IvanKryzhanovskyi:
## Refs
https://issues.folio.org/browse/UIDATIMP-933


KaterynaSenchenko (approved):

zburke (commented):
Removing support for an interface (`source-storage-records` `2.0`) requires updating the major version of this package. Alternatively, if this module remains compatible with `2.0` and adds support for `3.0` you can change the dependency to `2.0 3.0` and bump the module's minor-version (since that hasn't already happened, despite new commits since the last release). 

Please fix this error before the next release.

PR #1097, opened 2022-04-13
BF-263 avoid nested Panesets in local SearchAndSort

zburke:
Nested `<Paneset>`s have some wacky CSS side-effects that are difficult
to understand in their entirety. It would be good to untangle that, but
avoiding nesting may be the easiest short-term solution.

Refs [BF-263](https://issues.folio.org/browse/BF-263)


zburke:
@IvanKryzhanovskyi, @JohnC-80 , with the extra `<Paneset>` removed from the local `SearchAndSort` component in ed1ed49, I cannot get the following tests to pass:

* ./src/settings/ActionProfiles/ActionProfiles.test.js
* ./src/settings/MappingProfiles/tests/MappingProfiles.test.js
* ./src/settings/FileExtensions/FileExtensions.test.js
* ./src/settings/MatchProfiles/MatchProfiles.test.js
* ./src/settings/JobProfiles/tests/JobProfiles.test.js
* ./src/components/UploadingJobsDisplay/UploadingJobsDisplay.js
* ./src/components/SearchAndSort/SearchAndSort.test.js

I think the source of the trouble is a `<Layer>` that expects to be wrapped in a `<Paneset>`. I tried mocking that locally in `FileExtensions.test.js`:
```
jest.mock('@folio/stripes/components', () => ({
  ...jest.requireActual('@folio/stripes/components'),
  Layer: jest.fn(({ children }) => <>{children}</>),
}));
``` 
but that generates a complaint about connect:
> Could not find “store” in the context of “Connect(FileExtensions)“. Either wrap the root component in a <Provider>, or pass a custom React context provider to <Provider> and the corresponding React context consumer to Connect(FileExtensions) in connect options.

Adding that wrapper, however, generates a complaint from final form: 
> No onSubmit function specified

(FileExtensions.test calls `renderWithReduxForm` instead of `renderWithFinalForm`, which I think was maybe just an error, but regardless the situation remains the same: when the `store` provider is present there's an `onSubmit` error but when it's absent there's a `missing store` error.)

I'm stumped at this point. It seems clear to me that these failures are just test failures, related to how the redux and Paneset contexts are built up (or not) in the test environment, rather than true indications of failure, but I can't figure out how to get the test env correctly configured.

Javokhir12 (approved):

PR #1105, opened 2022-04-28
UIDATIMP-1147: For the Data Import View all page, change from Load more to Paginated

OleksandrHladchenko1:
## Purpose
 For the Data Import View all page, change from Load more to Paginated

## Approach
 
* Replace **click** with **prev-next** in pagingType prop in SearchAndSort component
*  Filtered empty records for job execution's filter 

## Refs
 https://issues.folio.org/browse/UIDATIMP-1147

## Screenshots

https://user-images.githubusercontent.com/85172747/165781644-54c7164a-3da3-4276-b1ac-45db2afd8f00.mp4


src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -414,6 +425,7 @@ class ViewAllLogs extends Component {
 
     const resultsFormatter = this.getResultsFormatter();
     const columnMapping = getJobLogsListColumnMapping({ isAllSelected, handleSelectAllCheckbox });
+    const itemToView = sessionStorage.getItem('job-logs.position');
```
Javokhir12, line 428: @OleksandrHladchenko1 I think `storage key` should be the same as `onMarkPosition`  and `resetMarkedPosition` handlers. Also, `itemToView` prop expects `object` so we should parse it.
```suggestion
    const itemToView = JSON.parse(sessionStorage.getItem('@folio/data-import.position'));
```

mariia-aloshyna (approved):

Javokhir12 (approved):

PR #1113, opened 2022-05-11
UIDATIMP-1146-FIX: Remove extra timeout and import pagingTypes correctly

Javokhir12:
## Purpose

- Remove extra timeout and import `pagingTypes` correctly
## Approach

- Remove extra timeout and import `pagingTypes` correctly


Javokhir12:
@mariia-aloshyna as you can see the reason why I added timeout is that it is failing with `Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error.`, but it is passing locallly.

mariia-aloshyna:
Hi @Javokhir12 re-running jobs usually fixes the problem

Javokhir12:
> Hi @Javokhir12 re-running jobs usually fixes the problem

I see, thanks for the tip.

PR #1136, opened 2022-05-24
 UIDATIMP-1166: Store job profile tree data in Redux store instead of sessionStorage

mariia-aloshyna:
## Purpose

User is getting error 'Something Went Wrong' on the UI when opening a data import profile for editing. This issue is not happening for all the data import profiles and apparently seems to be happening for profiles having more data/conditions.

As per the error , we could see that json string object is being put in session storage but the system is failing to do so as the session storage for most browsers has a limit of 5 MB. It seems that it is due to setting up of below API response in session storage. 

## Approach

- Store job profile tree data in Redux store instead of sessionStorage

## Refs

https://issues.folio.org/browse/UIDATIMP-1166


Javokhir12 (approved):

src/components/ProfileTree/ProfileBranch.js
```diff
@@ -3,13 +3,19 @@ import React, {
   useEffect,
   useState,
 } from 'react';
+import {
+  useDispatch,
+  useSelector
```
OleksandrHladchenko1, line 8:
```suggestion
  useDispatch,
  useSelector,
```

src/components/ProfileTree/ProfileTree.js
```diff
@@ -1,21 +1,34 @@
 import React, {
   memo,
+  useEffect,
   useState,
 } from 'react';
+import {
+  useDispatch,
+  useSelector
```
OleksandrHladchenko1, line 8:
```suggestion
  useDispatch,
  useSelector,
```

src/components/ProfileTree/ProfileTree.js
```diff
@@ -1,21 +1,34 @@
...
   last,
   snakeCase,
+  get,
+  set,
 } from 'lodash';
 import classNames from 'classnames';
 
+import {
+  setCurrentProfileTreeContent,
+  setProfileTreeContent
```
OleksandrHladchenko1, line 24:
```suggestion
  setCurrentProfileTreeContent,
  setProfileTreeContent,
```

src/settings/JobProfiles/JobProfilesForm.js
```diff
@@ -40,13 +45,19 @@ import {
...
   PROFILE_LINKING_RULES,
   ASSOCIATION_TYPES,
   PROFILE_TYPES,
   LAYER_TYPES,
   isFieldPristine,
 } from '../../utils';
+import {
+  clearCurrentProfileTreeContent,
+  clearProfileTreeContent,
+  setProfileTreeContent
```
OleksandrHladchenko1, line 59:
```suggestion
  clearCurrentProfileTreeContent,
  clearProfileTreeContent,
  setProfileTreeContent,
```

OleksandrHladchenko1 (approved):

PR #1161, opened 2022-06-15
UIDATIMP-1156: When closing the individual job log, return to the DI Landing page or View all properly

Javokhir12:
## Purpose
- When closing the individual job log, return to the DI Landing page or View all properly

## Approach
- Use `sessionStorage` to persist `selected logs`.
- Pass state to `<Link />` to return to correct page from the individual job log

## Refs
- https://issues.folio.org/browse/UIDATIMP-1156

## Screenshots

https://user-images.githubusercontent.com/89512426/173779450-4d6c0198-541d-4523-bbbd-94af4fce9128.mp4


mariia-aloshyna:
Hi @Javokhir12 I'm wondering if we can use **sessionStorage** to save checkboxes` state instead of **Redux**? The thing is that redux is mostly used to manage the state of the app and we need just to "remember" checkboxes' state, so I guess we could get by with sessionStorage. What do you think?

Javokhir12:
Hi @mariia-aloshyna. I was following other module's pattern but anyways good point. I updated the code.

mariia-aloshyna (approved):

jakhongirabdukhamdiov (approved):

PR #1164, opened 2022-06-21
UIDATIMP-1127: Use API for deleting data import logs on Data Import view all page

mariia-aloshyna:
## Purpose

Use API for deleting data import logs once the endpoint is ready on the Data Import View all, so that users can delete one or more logs.

## Approach

- Call `deleteJobExecutions` function to delete logs
- Update `mutator.query` to force `shouldRefresh` method to fetch fresh logs list

## Refs

https://issues.folio.org/browse/UIDATIMP-1127

## Screenshots

chrome_5iD2sTMjEr


Javokhir12 (approved):

Javokhir12:
Hi @mariia-aloshyna, what do you say if we disable all chekboxes not only `check all`, it would be the same as landing page.

jakhongirabdukhamdiov (approved):

PR #1175, opened 2022-07-01
UIDATIMP-1199: fix Checkbox on page 2+ of View all log list does not work properly

jakhongirabdukhamdiov:
## Purpose
Checkbox on page 2+ of View all log list does not work properly

## Approach
Before checking is all logs selected. remove extra empty values. 

## Refs
[UIDATIMP-1199](https://issues.folio.org/browse/UIDATIMP-1199)

## Screenshots
https://user-images.githubusercontent.com/100832608/176848060-384e42aa-786e-4857-b858-8a4b2f743e26.mp4

## Comment
On pagination pages other than the first one, the deleted logs are not removed from UI immediately but the network tab shows a get request result with updated logs, but the component did update and shows the previous records for both when checking equality.


src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -411,13 +408,19 @@ class ViewAllLogs extends Component {
     } = this.props;
     const { isLogsDeletionInProgress } = this.state;
 
+    let isAllSelectedByValues = isAllSelected;
```
Javokhir12, line 411: @jakhongirabdukhamdiov what does `isAllSelectedByValues` mean? Could you come up more meaningful name?

src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -411,13 +408,19 @@ class ViewAllLogs extends Component {
...
+    if (isAllSelected) {
+      const nonEmptyRecords = resources.records.records.map(record => record.id).filter(Boolean);
```
Javokhir12, line 413: Could you change this part in manifest something like this:
```suggestion
      const nonEmptyRecords = resources.jobLogs.records.map(record => record.id).filter(Boolean);
```

src/routes/ViewAllLogs/ViewAllLogs.js
jakhongirabdukhamdiov: in the checkboxlist component you can see that the selected records are stored in session storage and decides isAllSelected by the length of the values that are stored and current records of the new page. So when we change from one pagination page to the next it shows it's all selected if next page also has the same number of records, not considering the values that are stored, that's the main reason I chose the name of byValues.

src/routes/ViewAllLogs/ViewAllLogs.js
jakhongirabdukhamdiov: already done with jobExecutions.

OleksandrHladchenko1 (approved):

src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -322,21 +323,27 @@ class ViewAllLogs extends Component {
...
-
+      refreshRemote(this.props);
       deselectAll();
       this.hideDeleteConfirmation();
       this.showDeleteLogsSuccessfulMessage(jobExecutionDetails.length);
+
+      if (resultOffset !== 0) {
+        mutator.resultOffset.replace(resultOffset - RESULT_COUNT_INCREMENT);
+        setTimeout(() => mutator.resultOffset.replace(resultOffset));
+      }
```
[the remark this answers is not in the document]
mariia-aloshyna: Hi @Javokhir12 thank you a lot!

PR #1176, opened 2022-07-04
UIDATIMP-1201: Data import Reset all bug

Javokhir12:
## Purpose
- When users reset the Data import View all page, the default sort is lost

## Approach
- pass `onResetAll` prop to `SearchAndSort` to reset sort to the initial value.

## Refs
- https://issues.folio.org/browse/UIDATIMP-1201


src/routes/ViewAllLogs/ViewAllLogs.js
```diff
@@ -396,6 +396,17 @@ class ViewAllLogs extends Component {
...
   }
 
+  handleResetAll = () => {
+    const { mutator: { query: { update } } } = this.props;
+
+    update({
+      filters: '',
+      sort: '-completedDate',
+      query: '',
+      qindex: '',
```
mariia-aloshyna, line 406: Can this be moved to const and used here and in manifest for `query: { initialValue }`?

mariia-aloshyna (approved):

PR #1193, opened 2022-07-29
UIDATIMP-1223: Change options for invoice level adjustments in Invoice field mapping

Javokhir12:
## Purpose
- When `Pro rate` field has `Not prorated` value, hide `Separate from` value from `Relation to total` field list.
- When `Relation to total` field has `Separate from` value, hide `Not prorated` value from `Pro rate` list.
- Also, I found a bug with `Add fund distribution button`. The problem is that when there are more than 1 `Invoice adjustments` and one of them enable `Add fund distribution button`, other `adjustments`'s buttons will be enabled even though others don't enable their buttons.

## Approach
- Remove shared state `isFundDistribution` which caused shared state problem
- Scope field values and fund distribution button logic in `renderAdjustment` function.

## Refs
- https://issues.folio.org/browse/UIDATIMP-1223

### Add fund distribution button bug

https://user-images.githubusercontent.com/89512426/181759221-d027dd85-c295-4aea-baac-82ae68132c2b.mp4

### After

https://user-images.githubusercontent.com/89512426/181759713-e1e70e90-4a18-44d6-a67b-7ed64a5be84c.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Bugs fixed:
 * View all logs: filters are not updated after logs deletion (UIDATIMP-1219)
 * View all logs: User and Job filter contains only 10 records (UIDATIMP-1220)
+* Change options for invoice level adjustments in Invoice field mapping (UIDATIMP-1223)
 
 ## [5.2.1](https://github.com/folio-org/ui-data-import/tree/v5.2.1) (2022-07-27)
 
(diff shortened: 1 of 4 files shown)


OleksandrHladchenko1 (approved):

mariia-aloshyna (approved):

mariia-aloshyna:
Hi @Javokhir12 I see that Add fund distribution button becomes disabled when Pro rate value is not equal to Not prorated, is this the expected behavior?

Javokhir12:
> Hi @Javokhir12 I see that Add fund distribution button becomes disabled when Pro rate value is not equal to Not prorated, is this the expected behavior?

Hi @mariia-aloshyna Yes it was the default before I made changes. If you read the text at the above the button it says: `Include field mappings for fund distribution only if Pro rate = Not prorated`

PR #1217, opened 2022-09-16
UIDATIMP-1235: UI flow for import jobs that finish before a user can cancel

OleksandrHladchenko1:
## Purpose

* To provide a clearer message when a user starts to cancel a job, but does not select an option on the cancellation modal before the import job finishes.
* As a staff person working with an import that needs to be cancelled
I want a clear indication if an import job finishes before I have replied to the confirmation modal
So that I will not be confused by what happened

## Approach

* Store runinng jobs' hrIds is `redux store`
* When job appears in the logs, its hrId is removed from `store`
* When new job appears in `Running` section, its hrId is added to `store`
*  When select job for stopping, and then it shows in the logs, `ConfirmationModal` is closed and `ErrorModal` is shown
* Made `Job` component as functional instead of class component

## Refs

[UIDATIMP-1235](https://issues.folio.org/browse/UIDATIMP-1235)

## Screenshots

https://user-images.githubusercontent.com/85172747/190679585-cab3f79f-883a-48cf-95df-0c04da07fb6a.mp4


mariia-aloshyna:
@folio-org/fe-tl-reviewers please, take a look

translations/ui-data-import/en_US.json
```diff
@@ -900,5 +900,7 @@
     "modal.cancelRunningJob.message": "Are you sure that you want <b>to cancel this import job?</b><p><b>Note:</b> Cancelled jobs cannot be restarted. Records created or updated before{break}the job is cancelled cannot yet be reverted.</p>",
     "modal.cancelRunningJob.cancel": "No, do not cancel import",
     "modal.cancelRunningJob.confirm": "Yes, cancel import job",
+    "modal.confirmJobFinished.header": "Import job finished before cancellation",
```
vashjs, line 903: Please check en.json for the same translations. I'm not sure if the rest of the files will be populated based on en_US.json

src/routes/Home.js
```diff
@@ -92,6 +126,20 @@ export class Home extends Component {
 
     setList(logs);
     this.setState({ logs });
+    this.onLogsUpdate(logs);
+  }
+
+  onLogsUpdate = logs => {
+    for (let i = 0; i < logs.length; i++) {
+      if (this.props.hrIds.includes(logs[i].hrId)) {
```
vashjs, line 134: `logs[i].hrId` can be moved to variable, used 3 times

vashjs (approved):
Please add line in CHANGELOG.md if it's required.

PR #1241, opened 2022-10-31
UIDATIMP-1290: Invoice field mapping profile: Vendor name lookup is hidden

mariia-aloshyna:
## Purpose

Vendor name lookup is hidden when the field is disabled but the field is required so it's unable to save the profile

## Approach

- Show `Organization look-up` even if the text field is disabled, otherwise, I do not see the need to use this component at all, you can simply render the disabled text field
- Add possibility to clear the field value even if the field is disabled

## Refs

https://issues.folio.org/browse/UIDATIMP-1290

## Screenshots

image


OleksandrHladchenko1 (approved):

mariia-aloshyna:

Terala-Priyanka (approved):

PR #1245, opened 2022-11-01
UIDATIMP-1265: Add required field asterisks & validation action to Order field mapping profile

mariia-aloshyna:
## Purpose

To add MARC mapping validation and indicate required fieds

## Approach

- Wrap fields into `WithValidation` component and add validation to the fields
- Add validation for required fields
- Add unit tests

## Refs

https://issues.folio.org/browse/UIDATIMP-1265

## Screenshots

image
image
image
image


OleksandrHladchenko1 (approved):

mariia-aloshyna:

zburke (commented):
The changes here are much more expansive than what is specified in the ticket ("indicate required fields") or in the PR description. That said, I believe the refactoring is all in the spirit of making validation more consistent across the board, which is a good thing. 

It would be nice to see unit tests indicating the same things shown in the description PRs, namely:

1. Required fields show a warning after blur without a value.
2. The form's submit handler is not invoked when the form is submitted but required values are missing.
3. MARC-validated-fields show a validation error when they contain an un-quoted, non-MARC value.
4. MARC-validated-fields do not show a validation error when they contain a quoted, non-MARC value.
5. MARC-validated-fields do not show a validation error when they contain an un-quoted MARC value.

Finally, the first screenshot here shows only HTML validation being used to handle required fields. I so so wish it were enough, but HTML-only validation just doesn't work consistently across platforms/browsers, etc. In other apps ([e.g. from ui-users](https://github.com/folio-org/ui-users/blob/72af6c2185c85c01591646c21aa50e72a018212c/src/views/UserEdit/UserForm.js#L50-L52)) we include required fields in a form-level validation function. Is there something similar in play here where we can add these required fields?

mariia-aloshyna:
> 

Hi @zburke I added a form-level `required` validation to the fields.
Also, in the scope of the ticket we decided to implement also MARC validation (it's split from the main story, I added comment there, we just need to update the description for the current ticket)
I added some unit tests for each of the sections but I prefer to implement the form validation in the other story for the whole form (UIDATIMP-1291)
What do you think?

zburke (approved):
> I added some unit tests for each of the sections but I prefer to implement the form validation in the other story for the whole form (UIDATIMP-1291)

This makes sense, but I am always wary of postponing tests. There is a lot happening in this PR, and it's the passing tests that provide confidence that everything continues to operate correctly. If test coverage erodes, then big changes like this will feel riskier.

PR #1249, opened 2022-11-10
UIDATIMP-1285: After duplicating profiles the previous view details information is displayed

OleksandrHladchenko1:
## Purpose

* After duplicating profiles duplicated view details information is displayed

## Approach

* Added `shouldRefresh` field in the manifest file for every profile type

## Refs

[UIDATIMP-1285](https://issues.folio.org/browse/UIDATIMP-1285)

## Screenshots

https://user-images.githubusercontent.com/85172747/201123906-305da7a3-d5ae-4f88-a62e-16b0641453c2.mp4


zburke (commented):
The [documentation indicates](https://github.com/folio-org/stripes-connect/blob/9ec640511ec17b7aa38be721ecc5653fde0cdce1/doc/api.md) `shouldRefresh` is used "to indicate if the given resource should be refreshed when another resource is mutated". IOW, all this will do is prevent a resource from being re-fetched. 

Can you explain why this fix works?

OleksandrHladchenko1:
Hi @zburke When I duplicate profile, View pane fetches duplicated record, according to its `manifest`. But Edit form, after saving profile, according to its `manifest`, fetches previous profile right after View pane, so previous one is displayed. `shouldRefresh` don't let the Edit form to fetch the old profile.

mariia-aloshyna (approved):

zburke:
Thanks @OleksandrHladchenko1, that's a helpful explanation. IIUC, the problem is really that both the edit and view pages use the same resource name, e.g. 
```
$ grep -r 'actionProfile: {' ./src | grep -v test
./src/settings/ActionProfiles/ViewActionProfile/ViewActionProfile.js:    actionProfile: {
./src/settings/ActionProfiles/EditActionProfile/EditActionProfile.js:  actionProfile: {
```
and since stripes-connect resources are global :( the edit-resource overwrites the view resource. Do you think passing `shouldRefresh` is truly the best option, or would using distinct resource names, e.g. `actionProfile` and `actionProfileEdit`, be a better way to avoid this unpleasant side-effect? I understand that your fix is very simple and it works, but I worry that it's a bit opaque. If you want to stick with the `shouldRefresh` approach, please add comments to the manifests to explain why that setting is necessary here since it's not normally required, and make sure the explanation gets into the git commit when you squash-and-merge.

PR #1253, opened 2022-11-14
UIDATIMP-1300: Data import settings: Shortcuts for creating, editing and duplicating don't work

OleksandrHladchenko1:
## Purpose

* Enable using shortcuts for creating, editing and duplicating profiles

## Approach

* When we try to create new profile, basic path is splitted by `/`, cut first four elements (`'', 'settings', 'data-import', 'profileType'`) and push the layer type `create` and return joined path with search query (`/settings/data-import/profileType/create?sort=name`)
* When we try to edit or duplicate profile, basic path is splitted by `/` (`'', 'settings', 'data-import', 'profileType', 'view', 'profileId'`) we look for the view layer and replace it with edit or duplicate, return joined path with search query (`/settings/data-import/profileType/edit/profileId?sort=name`)

## Refs

[UIDATIMP-1300](https://issues.folio.org/browse/UIDATIMP-1300)

## Screenshots

https://user-images.githubusercontent.com/85172747/201654183-a1dcb1d1-1a74-40af-8c97-0220c4c640bd.mp4

### DIFF
--- a/src/utils/craftLayerURL.js
+++ b/src/utils/craftLayerURL.js
@@ -12,15 +12,15 @@ export const createLayerURL = (location, type) => {
     pathname,
     search,
   } = location;
-  const splittedPath = pathname.split('/');
+  const pathList = pathname.split('/');
 
   if (type === LAYER_TYPES.CREATE) {
-    splittedPath.length = 4;
-    splittedPath.push(type);
+    pathList.length = 4;
+    pathList.push(type);
   } else {
-    const itemToChangeIndex = splittedPath.findIndex(item => item === LAYER_TYPES.VIEW);
-    splittedPath[itemToChangeIndex] = type;
+    const itemToChangeIndex = pathList.findIndex(item => item === LAYER_TYPES.VIEW);
+    pathList[itemToChangeIndex] = type;
   }
 
-  return `${splittedPath.join('/')}${search}`;
+  return `${pathList.join('/')}${search}`;
 };
--- a/src/utils/craftLayerURL.js
+++ b/src/utils/craftLayerURL.js
@@ -3,24 +3,19 @@ import { LAYER_TYPES } from './constants';
 /**
  * Description: TBD
  *
- * @param {object} location comes from withRouter
- * @param {string} type layer type
+ * @param {object} baseUrl base location url
+ * @param {string} layerType layer type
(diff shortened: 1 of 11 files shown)


src/utils/craftLayerURL.js
```diff
@@ -10,8 +12,15 @@ export const createLayerURL = (location, type) => {
     pathname,
     search,
   } = location;
+  const splittedPath = pathname.split('/');
 
-  const url = `${pathname}${search}`;
+  if (type === LAYER_TYPES.CREATE) {
+    splittedPath.length = 4;
```
ryandberger, line 18: This shouldn't be necessary. `.push()` increases the size of the array automatically. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

ryandberger, line 18: Maybe call this `pathList` instead? `splitted` isn't a recognized English word :)

OleksandrHladchenko1: Hi @ryandberger! 
I used `splittedPath.length = 4` to **cut first four elements from the path**. Because we can get path like this `/settings/data-import/profileType?sort=name` or `/settings/data-import/profileType/view/profileId?sort=name`.
And as the result we get array like this `['', 'settings', 'data-import', 'profileType']`.
Then I push the layer type `create` and join this array. As the result we get path `/settings/data-import/profileType/create?sort=name`

uladislausamets (approved):

src/utils/craftLayerURL.js
```diff
@@ -10,8 +12,15 @@ export const createLayerURL = (location, type) => {
     pathname,
     search,
   } = location;
+  const pathList = pathname.split('/');
 
-  const url = `${pathname}${search}`;
+  if (type === LAYER_TYPES.CREATE) {
+    pathList.length = 4;
```
ryandberger, line 18: This should not be necessary. `.push()` automatically increases the array size. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

OleksandrHladchenko1: @ryandberger Tried a new approach, please, review

mariia-aloshyna (approved):

PR #1303, opened 2023-01-06
UIDATIMP-1333: Improve mapping profile for Orders (follow-up)

mariia-aloshyna:
## Purpose

UPDATES after re-checking via new improvements from UI.

## Approach

- Set "repeatableFieldAction": "EXTEND_EXISTING" for `contributors` and `refNumbers` fields.
- Adjust sending values for "Create Inventory" fields.

## Refs

https://issues.folio.org/browse/UIDATIMP-1333

### DIFF
--- a/src/settings/MappingProfiles/detailsSections/constants.js
+++ b/src/settings/MappingProfiles/detailsSections/constants.js
@@ -7,8 +7,8 @@ export const TRANSLATION_ID_PREFIX = 'ui-data-import.settings.mappingProfiles.ma
 export const FIELD_NAME_PREFIX = 'profile.mappingDetails.mappingFields';
 
 export const CREATE_INVENTORY_TYPES = {
-  INSTANCE_HOLDINGS_ITEM: 'Instance, holdings, item',
-  INSTANCE_HOLDINGS: 'Instance, holdings',
+  INSTANCE_HOLDINGS_ITEM: 'Instance, Holding, Item',
+  INSTANCE_HOLDINGS: 'Instance, Holding',
   INSTANCE: 'Instance',
   NONE: 'None',
 };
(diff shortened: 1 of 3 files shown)


src/settings/MappingProfiles/detailsSections/constants.js
```diff
@@ -7,8 +7,8 @@ export const TRANSLATION_ID_PREFIX = 'ui-data-import.settings.mappingProfiles.ma
 export const FIELD_NAME_PREFIX = 'profile.mappingDetails.mappingFields';
 
 export const CREATE_INVENTORY_TYPES = {
-  INSTANCE_HOLDINGS_ITEM: 'Instance, holdings, item',
-  INSTANCE_HOLDINGS: 'Instance, holdings',
+  INSTANCE_HOLDINGS_ITEM: 'Instance, Holding, Item',
+  INSTANCE_HOLDINGS: 'Instance, Holding',
```
zburke, line 11: I have questions:
1. The Jira the _UI_ should have the values "Instance, Holding, Item"; "Instance, Holding"; "Instance"; "None". The fact that this change is taking place here rather than in `translations/ui-data-import/en.json` suggests these values are not correctly i18n'ed.
2. Are we confident about "holding" rather than "holdings" or "holdings record"? The [FOLIO glossary](https://wiki.folio.org/display/FOLIOtips/Glossary) uses "holdings". Additionally, the [FOLIO standard for text case](https://ux.folio.org/docs/guidelines/language-rules/text-case-rules/) indicates that only "Instance" should be capitalized here.

OleksandrHladchenko1 (approved):
Update CHANGELOG, please

src/settings/MappingProfiles/detailsSections/constants.js
mariia-aloshyna: Hi @zburke this const is values sending to BE, UI labels are rendered as translations, e.g. https://github.com/folio-org/ui-data-import/blob/master/src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/EResourcesDetails.js#L66-L80

mariia-aloshyna:
> Update CHANGELOG, please

Hi @OleksandrHladchenko1 it's follow-up changes, CHANGELOG has been already updated

zburke (approved):

PR #1304, opened 2023-01-10
UIDATIMP-1344: Error after selecting order option on field mapping profile

OleksandrHladchenko1:
## Purpose

* Remove "Something went wrong" error message after `Order` record type selection on field mapping profile 

## Approach

* Removed dependencies to prevent rerendering in `useEffect` hook in `OrderInformation` component
*  Cover `purchaseOrderLinesLimitValue` with `useMemo` hook

## Refs

[UIDATIMP-1344](https://issues.folio.org/browse/UIDATIMP-1344)

## Screenshots

https://user-images.githubusercontent.com/85172747/211541295-55dd2817-feba-430f-8fe0-1583b9b8e5f1.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Invoice field mapping profile: Vendor name lookup is hidden (UIDATIMP-1290)
 * Order and Order line field mapping: View screen updates (UIDATIMP-1304)
 * Order and Order line field mapping: Create/Edit screen updates (UIDATIMP-1305)
+* Error after selecting order option on field mapping profile (UIDATIMP-1344)
 
 ## [5.3.10](https://github.com/folio-org/ui-data-import/tree/v5.3.10) (2022-12-05)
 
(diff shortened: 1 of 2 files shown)


zburke (approved):
It looks like this PR essentially undoes some work from #1301, restoring things to how they operated previously. That's fine, but there's no real explanation of why this fix works aside from the hint provided by the eslint comment, i.e. that some deps were causing an infinite render loop. Lint warns about this stuff for a reason, and while it's fine to turn it off when it gets things wrong, but it's helpful to explain why for future devs who work on this code, e.g.
```
// ignore these deps: purchaseOrderLinesLimitSetting.records, setReferenceTables
// because they cause in infinite-render loop. We only care about `...hasLoaded`, 
// which indicates when our query has finished
```

The PR description also states

> Cover purchaseOrderLinesLimitValue with useMemo hook

but there is no such change here. 

Finally, the PR description/commit subject should state the solution, not the problem, so somebody reviewing commits can see what changed, e.g. sth like, "Avoid infinite rerender in OrderInformation during useEffect".

PR #1308, opened 2023-01-19
UIDATIMP-1348: Receiving value disappears from Order field mapping profile after default order line limit updated

mariia-aloshyna:
## Purpose

To display the actual `PO lines limit` value from Order Settings on the view and create/edit screen. To send the `Receiving workflow` field value wrapped in quotes.

## Approach

- Wrap the `Receiving workflow` field value in quotes before sending to BE
- Fetch and display `PO lines limit` value from Order Settings

## Refs

https://issues.folio.org/browse/UIDATIMP-1348

### DIFF
--- a/src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/POLineDetails.js
+++ b/src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/POLineDetails.js
@@ -1,7 +1,4 @@
-import React, {
-  useCallback,
-  useState,
-} from 'react';
+import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
 import {
   FormattedMessage,
@@ -42,7 +39,6 @@ import {
   BOOLEAN_ACTIONS,
   BOOLEAN_STRING_VALUES,
   validateMARCWithDate,
-  validateMARCWithElse,
 } from '../../../../../utils';
 
 export const POLineDetails = ({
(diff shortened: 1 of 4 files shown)


OleksandrHladchenko1 (approved):

src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/POLineDetails.js
```diff
@@ -112,27 +106,20 @@ export const POLineDetails = ({
 
   const onReceivingWorkflowParse = useCallback(
     value => {
-      setReceivingWorkflow(value);
-      if (value === `"${RECEIVING_WORKFLOW.INDEPENDENT}"`) return BOOLEAN_STRING_VALUES.TRUE;
-      if (value === `"${RECEIVING_WORKFLOW.SYNCHRONIZED}"`) return BOOLEAN_STRING_VALUES.FALSE;
+      if (value === `"${RECEIVING_WORKFLOW.INDEPENDENT}"`) return `"${BOOLEAN_STRING_VALUES.TRUE}"`;
```
vashjs, line 109: @mariia-aloshyna RECEIVING_WORKFLOW.INDEPENDEN is it not string? Not sure that we need wrap it with quotes.
`if (value === RECEIVING_WORKFLOW.INDEPENDENT) return BOOLEAN_STRING_VALUES.TRUE;`

The same questuin for all other places. Maybe we can create some kind of map here, for example:
```
const map = {
 [RECEIVING_WORKFLOW.INDEPENDENT]: BOOLEAN_STRING_VALUES.TRUE,
 [RECEIVING_WORKFLOW.SYNCHRONIZED]: BOOLEAN_STRING_VALUES.FALSE
// all other fields
}
```

we can avoid all of these IF statements.

vashjs (approved):

src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/POLineDetails.js
vashjs: Thanks for clarification

PR #1311, opened 2023-01-24
UIDATIMP-1353 Bump `stripes-*` and upgrade `react-redux` to `v8`

ryandberger:


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=6.0.0"
+    "node": ">=14.0.0"
```
zburke, line 12: `engines` can be omitted; this is a browser module, not a nodejs module, so it `engines` is irrelevant at runtime. I would vote for leaving it alone (and fixing this in a separate PR) or removing it altogether.

zburke (approved):
LGTM

OleksandrHladchenko1 (approved):

PR #1330, opened 2023-02-03
UIDATIMP-1368: Order field mapping profile: Show the expense class and code when default expense class is selected

mariia-aloshyna:
## Purpose

For the Order field mapping profile's Expense class field, when displaying the values in the dropdown list or in the field mapping field, the expense class codes do not display - only the expense class names. This applies to the Create/Edit screen and the View screen for the Order field mapping profile.

## Approach

- Add `optionTemplate` to show the correct format of options.

## Refs

https://issues.folio.org/browse/UIDATIMP-1368

## Screenshots

image

image


OleksandrHladchenko1 (approved):

mariia-aloshyna:

PR #1348, opened 2023-02-23
UIDATIMP-1383: Override purchase order lines limit setting field allows saving fractional numbers

mariia-aloshyna:
## Purpose

Override purchase order lines limit setting field does not have validation and allows a user to save fractional numbers.

## Approach

- Add validation to check the integer in quotation marks.
- Add unit tests

## Refs

https://issues.folio.org/browse/UIDATIMP-1383

## Screenshots

image
image
image


OleksandrHladchenko1 (approved):

src/utils/tests/formValidator.test.js
```diff
@@ -1,184 +0,0 @@
-import { renderWithIntl } from '@folio/stripes-data-transfer-components/test/jest/helpers';
```
artem-blazhko, line 1: Could you clarify what is the reason to delete that file?

mariia-aloshyna: It's not deleted, but just renamed :)
This is a way how Github shows it

artem-blazhko (approved):

PR #1362, opened 2023-03-09
UIDATIMP-1387: Order field mapping profile: Fix the UI and error message for the fund "Value" field

mariia-aloshyna:
## Purpose

If an Order field mapping profile is created with a fund code "Value" field of 100, no error message is displayed, and the order import does not complete properly.

## Approach

- Adjust regexp to indicate if a subfield for marc mapping is required
- Make a subfield required for the `Value` field
- Show validation error when non-marc values are not wrapped into quotation marks
- Add unit tests

## Refs

https://issues.folio.org/browse/UIDATIMP-1387

## Screenshots

image
image
image


OleksandrHladchenko1 (approved):

OleksandrHladchenko1:

vashjs (approved):

PR #1371, opened 2023-03-21
UIDATIMP-1400: Order field mapping profile: Inactivate some fields when Order format is Electronic

OleksandrHladchenko1:
## Purpose

* If the Order format in the field mapping profile is `Electronic resource`, then disable some fields in Order field mapping profile form

## Approach

* `clearFieldValue` method clears fields, when they are disabled
* `clearSubfieldValue` method clears field in the subfield, when it is disabled

## Refs

[UIDATIMP-1400](https://issues.folio.org/browse/UIDATIMP-1400)

## Screenshots

https://user-images.githubusercontent.com/85172747/226631532-f31dc202-0fa8-4737-b45d-afb6e7398354.mp4


vashjs (approved):

ryandberger (approved):

src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/PhysicalResourceDetails.js
mariia-aloshyna, line 1: Fix warnings, please

OleksandrHladchenko1: Good catch! Thanks!

src/settings/MappingProfiles/detailsSections/utils.js
```diff
@@ -286,8 +286,23 @@ export const getMappingQueryFromValue = valueFromDetails => {
   return valueFromDetails?.substring(0, mappingEndPosition === -1 ? valueFromDetails.length : mappingEndPosition);
 };
 
-export const setFieldValue = ({ path, setReferenceTables, value, isDisabled }) => {
-  const reference = isDisabled ? '' : value;
+export const clearFieldValue = ({ paths, setReferenceTables, value = '', isDisabled }) => {
+  if (isDisabled) {
+    paths.forEach(path => setReferenceTables(path, value));
+  }
```
mariia-aloshyna, line 292: This function shouldn't check whether the field is disabled or not, it should be checked in the place where you call this function (in useEffect e.g.). Also, this function shouldn't accept `value` as arg, it should just clear the field value so set an empty string, or `null`, because if I pass some value to the function then it won't actually clear the field, but set some value to it.

src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/PhysicalResourceDetails.js
```diff
@@ -69,23 +69,41 @@ export const PhysicalResourceDetails = ({
...
+      isDisabled: isPhysicalDetailsDisabled,
+      setReferenceTables,
+    });
+    clearFieldValue({
+      paths: [`profile.mappingDetails.mappingFields[${PHYSICAL_RESOURCE_DETAILS_FIELDS_MAP.VOLUMES}].subfields`],
+      isDisabled: isPhysicalDetailsDisabled,
+      setReferenceTables,
+      value: [],
+    });
+  }, [isCreateInventoryDisabled, isPhysicalDetailsDisabled]); // eslint-disable-line react-hooks/exhaustive-deps
```
mariia-aloshyna, line 106: Split this into separate `useEffects` with different dependencies, because every time `isCreateInventoryDisabled` prop has changed but `isPhysicalDetailsDisabled` not, `physicalDetailsDisabledPaths` and `[profile.mappingDetails.mappingFields[${PHYSICAL_RESOURCE_DETAILS_FIELDS_MAP.VOLUMES}].subfields]` will also be changed, it's unnecessary re-rendering

mariia-aloshyna (approved):

PR #1375, opened 2023-03-24
UIDATIMP-1401: Order field mapping profile: Inactivate some fields when Order format is Physical or Other

mariia-aloshyna:
## Purpose

If the Order format in the field mapping profile is Physical or Other, then Import will error if electronic resource details are included. Inactivate those electronic resource fields in the field mapping profile, so that those details cannot be accidentally mapped or supplied in the order data.

## Approach

- Create `useDisabledOrderFields` hook to get indicators of disabled fields
- Disable and clear required fields

## Refs

https://issues.folio.org/browse/UIDATIMP-1401

## Screenshots

chrome_Qh21udizE8


src/settings/MappingProfiles/detailsSections/hooks/useDisabledOrderFields.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ *
+ * @returns {{dismissElectronicDetails: boolean, dismissCreateInventory: boolean, dismissPhysicalDetails: boolean}}
+ * an object with indicators of disabled fields
+ */
+export const useDisabledOrderFields = () => {
+  const [poStatus, orderFormat] = useFieldMappingFieldValue([PO_STATUS_FIELD, ORDER_FORMAT_FILED]);
+
+  const dismissCreateInventory = useMemo(() => poStatus === PO_STATUS.OPEN, [poStatus]);
```
vashjs, line 22: I sure that using useMemo on lines 22,23,24 costs more then optimization it provides.  May we remove it?

https://react.dev/reference/react/useMemo#should-you-add-usememo-everywhere

OleksandrHladchenko1 (approved):

vashjs (approved):

PR #1432, opened 2023-07-10
UIDATIMP-1461: Clear search field when switching between profile types

OleksandrHladchenko1:
## Purpose

* If there is a search in the profile settings search box, delete it if the user changes to a different type of profile

## Approach

* When component is mounted set `query: ''` in addition to sort
* Set default width 430px for `name` column because it shrinks when switch between profiles

## Refs

[UIDATIMP-1461](https://issues.folio.org/browse/UIDATIMP-1461)


mariia-aloshyna (approved):

OleksandrHladchenko1:
@folio-org/fe-tl-reviewers please, take a look

Dmytro-Melnyshyn (approved):

PR #1435, opened 2023-07-12
UIDATIMP-1473: Allow to support only text messages from BE in `NetworkMessage` component

OleksandrHladchenko1:
## Purpose

* Since BE sends error messages in plain text format (previously both plain text and code format e.g `jobProfile.duplication.invalid`), it is necessary to change behavior in `NetworkMessage` component.

## Approach

* If we get error message with plain text we'll show `errorMessage`
* If we get error without message we'll show `communicationProblemMessage`
* If we get success message we'll show `successMassage`
* Removed unused translations
* Added unit test

## Refs

[UIDATIMP-1473](https://issues.folio.org/browse/UIDATIMP-1473)

### Before
image

### After
image


src/components/Callout/NetworkMessage.js
```diff
@@ -14,30 +14,35 @@ export const NetworkMessage = memo(({
...
-            values={{
-              name: recordName || profile?.name || extension,
-              action: <FormattedMessage id={`ui-data-import.action.${action}`} />,
-            }}
-          />
-        )
-      }
+      {type === 'error' && !messageId && communicationProblemMessage}
+      {type === 'error' && messageId && errorMessage}
+      {type !== 'error' && sucessMassage}
```
mariia-aloshyna, line 45:
```suggestion
      {type !== 'error' && successMessage}
```

mariia-aloshyna (approved):

ryandberger (approved):

PR #1446, opened 2023-08-02
UIDATIMP-1463 add messaging to inform user of large file splitting, add backend config check to UploadJobContext

JohnC-80:
# Purpose

This PR adds a backend configuration check to the UploadJobContext so that it can render the UI/perform upload logic accordingly.
The first conditional use of this functionality is to render a message in the UI within the ConfirmationModal of the 'Run' command from the JobProfile view/ Actions dropdown.

## Approach

The configuration request is a one-time call in the UI and will return a boolean for `splitStatus`. It's achieved here using stripes-core's exported `useOkapiKy` HOC which can be used to easily make requests to okapi (it includes tokens in the header for the request.) `ky` is a library that wraps `fetch` functionality, easing the setup of requests.

As work on this project proceeds, the `utils/multiplartUpload.js` file will be expanded to contain logic for the direct S3 multipart  upload.


translations/ui-data-import/en_GB.json
```diff
@@ -219,7 +219,7 @@
     "editJobProfile": "Edit job profile",
     "run": "Run",
     "modal.jobProfile.run.header": "Are you sure you want to run this job?",
-    "modal.jobProfile.run.message": "You have selected the <b>\"{name}\"</b> job profile to <b>Run</b> the uploaded files",
```
ncovercash, line 222: I think you'll have to get these fixed in Lokalise, iirc it overwrites

ncovercash (approved):

src/components/UploadingJobsContextProvider/UploadingJobsContextProvider.js
```diff
@@ -33,9 +37,24 @@ export class UploadingJobsContextProvider extends Component {
...
+  componentDidMount = () => {
+    this.getUploadConfiguration();
+  }
+
+  getUploadConfiguration = async () => {
+    const { okapiKy } = this.props;
+    const { splitStatus } = await getStorageConfiguration(okapiKy);
+    this.setState({
+      uploadConfiguration: {
+        canUseObjectStorage: splitStatus
```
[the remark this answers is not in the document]
JohnC-80: Things shouldn't render until this is loaded.

translations/ui-data-import/en_GB.json
JohnC-80: These are just to cover me a little bit before Lokalize from master is pulled over. I think Lokalize updates based on what's in en.json.

PR #1448, opened 2023-08-04
UIDATIMP-1463 add logic to split files and upload to S3 storage

JohnC-80:
## Purpose
Upload files to object storage, if available, using a multipart upload.

## Approach
The conditional logic here leans on the backend configuration obtained in UIDATIMP-1463. Since this information is obtained asyncronously, logic for immediately triggering uploads in the constructor needed to be moved to componentDidUpdate in order to wait for the configuration to arrive.

All backend requests are performed via `OkapiKy`. 
The file is locally split via JS `Blob.split` API, iterating through uploads one segment at a time.
The logic has to

1. Initialize a multipart upload (and obtain a signed upload URL from S3 services for the first, possibly only part)
2. If any subsequent uploads are necessary, an additional signed upload URL will have to be obtained for subsequent parts.
3. As parts complete, their 'entity tag' (ETag) is accumulated so that it can ultimately be submitted to finalize the multipart upload and have S3 assemble the uploaded parts.

Upload progress is maintained across multiple parts via a tiny accumulator class instance, per-file.
Currently, the response from finalizing is created on the client to fulfill the UI's wishes of getting an 'dateUploaded' value.

TODO: testing!

### DIFF
--- a/src/utils/multipartUpload.js
+++ b/src/utils/multipartUpload.js
@@ -31,7 +31,6 @@ function uploadPart(part, url, partNumber, totalParts, fileKey, progressHandler,
     partUploaderXhr.open('PUT', url, true);
     partUploaderXhr.upload.onprogress = event => {
       const { loaded, total } = event;
-      // console.log(`${partNumber} of ${totalParts} upload progress: ${(loaded / total) * 100}%`);
       const newEvent = {
         ...event,
         loaded: progressAccumulator.totalProgress + event.loaded,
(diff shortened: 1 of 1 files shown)


src/utils/multipartUpload.js
```diff
@@ -1,5 +1,124 @@
...
+  ).json();
+}
+
+function uploadPart(part, url, partNumber, totalParts, fileKey, progressHandler, progressAccumulator) {
+  return new Promise((resolve, reject) => {
+    const partUploaderXhr = new XMLHttpRequest();
+    partUploaderXhr.open('PUT', url, true);
+    partUploaderXhr.upload.onprogress = event => {
+      const { loaded, total } = event;
+      // console.log(`${partNumber} of ${totalParts} upload progress: ${(loaded / total) * 100}%`);
```
mariia-aloshyna, line 34: Remove this console.log, please

src/utils/multipartUpload.js
```diff
@@ -1,5 +1,124 @@
...
+      const eTag = await uploadPart(chunk, _uploadURL, currentPartNumber, totalParts, fileKey, progressHandler, progressAccumulator);
+      eTags.push(eTag);
+      currentPartNumber += 1;
+      currentByte += CHUNK_SIZE;
+    } catch (error) {
+      errorHandler(fileKey);
+      break;
+    }
+  }
+  // const finishResponse = await finishUpload(eTags, _uploadKey, _uploadId, ky);
```
mariia-aloshyna, line 110: Remove this comment if don't need it, please

mariia-aloshyna (approved):

ncovercash (approved):

PR #1453, opened 2023-08-16
UIDATIMP-1466 Add composite job details to Job component

JohnC-80:
[UIDATIMP-1466](https://issues.folio.org/browse/UIDATIMP-1466)
and
[UIDATIMP-1464](https://issues.folio.org/browse/UIDATIMP-1464)

## Purpose
Jobs-in-progress cards present data regarding the progress of a running job. With the addition of file-splitting/submission of multiple jobs, we consolidate multiple child job entities to a single card that displays the number of file slices remaining to be processed, the number of finished slices, errored slices, and canceled/discarded slices.

image

The job log display is also rendered with a 'Job parts' column when the configuration is active.

image

## Approach

The query needed to be adjusted to include 'subordinationTypeNotAny=COMPOSITE_CHILD' to pull down the composite job, as well as 'subordinationTypeNotAny=COMPOSITE_PARENT' for the job logs listing.
If the 'job' data for the card contains a `CompositeDetails` field, it will render the extra progress information.

The trickiest part is probably varying the query parameters for the jobs/logs request in `<DataFetcher/>`...
It makes use of `stripes-connect`'s functional path capability and waits for a return from the configuration call before requesting the jobs/logs. For composite data, an additional query parameter is required : `subordinationTypeNotAny`

##TO-DO 
Update/add tests for job component...

### DIFF
--- a/translations/ui-data-import/en_GB.json
+++ b/translations/ui-data-import/en_GB.json
@@ -17,6 +17,7 @@
     "jobProfileName": "Job profile",
     "jobExecutionHrId": "ID",
     "jobCompletedDate": "Ended running",
+    "jobParts": "Job parts",
     "noPreviewsJobsMessage": "No previews to show",
     "noRunningJobsMessage": "No running jobs to show",
     "loading": "Loading",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/data-import",
-  "version": "6.0.8",
+  "version": "6.0.10",
   "description": "Data Import manager",
   "main": "src/index.js",
   "repository": "folio-org/ui-data-import",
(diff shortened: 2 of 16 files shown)


translations/ui-data-import/en_GB.json
```diff
@@ -17,6 +17,7 @@
     "jobProfileName": "Job profile",
     "jobExecutionHrId": "ID",
     "jobCompletedDate": "Ended running",
+    "jobParts": "Job parts",
```
ncovercash, line 20: Do these do anything? I was under the impression that `en.json` was the only one that mattered and the others were all handled by Lokalise

ncovercash (approved):
Looks good to me, just sort out CI/testing issues!

translations/ui-data-import/en_GB.json
JohnC-80: en_GB may not... let me remove it and try... en_US may be the only one that gets used here.

ncovercash: i think it gets used in local stripes dev, but i thought it was overwrited with translation files — was just curious, no worries!

ncovercash (approved):
oops,just remembered you asked me to re-review! LGTM

PR #1455, opened 2023-08-18
UIDATIMP-1469 Handle cancelation of running composite job...

JohnC-80:
[UIDATIMP-1469](https://issues.folio.org/browse/UIDATIMP-1469)

This implements the cancellation action that's available on import cards. The layout and strings of the content of the following confirmation modal has been adjusted. For the most part, the existing logic path in the code has been maintained.

image


translations/ui-data-import/en_US.json
mariia-aloshyna, line 1: Aren't all other files except `en.json` updated by a bot?

src/components/Jobs/components/Job/Job.js
```diff
@@ -38,6 +39,7 @@ import { jobExecutionPropTypes } from './jobExecutionPropTypes';
 
 import * as API from '../../../../utils/upload';
 import * as CompositeJobFields from '../../../../utils/compositeJobStatus';
+import { cancelMultipartJob } from '../../../../utils';
```
ncovercash, line 42: Consolidate/fix these imports, there's multiple impotrs from `utils` and it seems like some are weirdly indented?

translations/ui-data-import/en_US.json
ncovercash: Looks like [yes](https://wiki.folio.org/display/I18N/How+to+Translate+FOLIO); I think John made those changes so the updated translations would show in his local build

ncovercash (changes requested):
Get that test coverage up and we should be good to go!

translations/ui-data-import/en_US.json
JohnC-80: I'm aware that translation files outside of en.json do get overwritten - We also need these translations to show up in our rancher environment pointed at this branch, and the lokalize process only runs against the master branch.

mariia-aloshyna (approved):

translations/ui-data-import/en.json
```diff
@@ -1124,6 +1131,7 @@
   "modal.cancelUpload.cancel": "No, do not delete",
   "modal.cancelUpload.confirm": "Yes, delete",
   "validation.enterValue": "Please enter a value",
+  "jobSummarySourceFileLabel": "Source file: ",
```
ncovercash, line 1134: Where is this being used?

ncovercash (approved):

PR #1458, opened 2023-08-21
UIDATIMP-1467 display jobParts column in Jobs using this profile section

JohnC-80:
Displays the "Job parts" column in the data-import settings Job Profile detail view.
[UIDATIMP-1467](https://issues.folio.org/browse/UIDATIMP-1467)

image

## Approach
This column is conditionally displayed based on the objectStorage capability/multipart job splitting. I moved the settings view into the children of the context provider to receive the value from there rather than rendering multiple, separate context providiers.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -42,7 +42,9 @@ class DataImport extends Component {
     if (showSettings) {
       return (
         <CommandList commands={keyboardCommands}>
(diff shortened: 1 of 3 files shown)


mariia-aloshyna (approved):

ncovercash (approved):
lgtm

PR #1463, opened 2023-09-01
UIDATIMP-1384: Provide "required" field for notes and electronic access subfields in field mapping

mariia-aloshyna:
## Purpose

To avoid hardcoding uri as required in di-core use the new "required" field on UI in mapping rule for uri.
Prevent creating empty notes.

## Approach

- Add `required` prop to initial values of mentioned subfields.

## Refs

https://issues.folio.org/browse/UIDATIMP-1384


zburke (approved):
True confession, I am not familiar enough with DI to understand if this implementation is correct. It appears to do what was asked in the Jira, so I am approving it based on that alone.

Dmytro-Melnyshyn (approved):

PR #1478, opened 2023-10-11
UIDATIMP-1540: Fetch Job summary data with for a particular tenant id

mariia-aloshyna:
## Purpose

To provide marc record and instance on the member tenant log page with jsons when marc-to-marc matching and marc-bib update has been performed on the central tenant by data import initiated on a member tenant.

## Approach

- Use custom `useTenantKy` hook to extend headers with tenantId

## Refs

https://issues.folio.org/browse/UIDATIMP-1540


src/hooks/useTenantKy.js
zburke, line 1: Do you think this hook belongs here in DI or could/should we just tweak `useOkapiKy` in stripes-core to accomodate requests that want to override the `tenant` value from `stripes`?

src/routes/ViewJobLog.js
```diff
@@ -58,16 +58,16 @@ export const ViewJobLog = () => {
...
-  const { data: instancesData } = useInventoryInstancesByIdQuery(instancesIds);
-  const { data: holdingsData } = useInventoryHoldingsByIdQuery(holdingsIds);
-  const { data: itemsData } = useInventoryItemsByIdQuery(itemsIds);
-  const { data: orderData } = useOrderByIdQuery(orderId);
-  const { data: poLinesData } = usePOLinesByIdQuery(poLinesIds);
-  const { data: invoicesData } = useInvoicesByIdQuery(invoicesIds);
-  const { data: invoiceLineData } = useInvoiceLineByIdQuery(invoiceLineId);
-  const { data: authoritiesData } = useAuthoritiesByIdQuery(authoritiesIds);
-  const { data: locationsData = [] } = useLocationsQuery();
+  const { data: srsRecordData } = useSRSRecordQuery(recordId, { tenantId: jobLogData?.sourceRecordTenantId});
```
OleksandrHladchenko1, line 61: Fix This warning please

src/hooks/useSRSRecordQuery.js
```diff
@@ -1,17 +1,16 @@
...
 
-import {
-  useNamespace,
-  useOkapiKy,
-} from '@folio/stripes/core';
+import { useNamespace } from '@folio/stripes/core';
 
-export const useSRSRecordQuery = recordId => {
-  const ky = useOkapiKy();
+import { useTenantKy } from '../hooks';
```
OleksandrHladchenko1, line 5: Fix This warning please

OleksandrHladchenko1 (approved):

zburke (approved):
LGTM, but what do you think about just implementing `useTenantKy` in `useOkapiKy` directly in stripes-core? `useTenantKy` feels extraneous given that `tenantId` is implemented as an optional prop. 

Aside, note that this value is `tenant` rather than `tenantId` in the rest of the stripes codebase and in `stripes.config.js`. Is it really named `tenantId` in the `jobLogData` responses? If so, that is exceptionally exceptionally unfortunate. From the stripes point of view, I'd say we should call it `tenant` in the hook and rename it to `tenant` when passing it into all those `use...Query()` hooks, but that introduces inconsistency with the property on those data structures. Where do you think consistency is most important?

src/hooks/useTenantKy.js
mariia-aloshyna: Regarding renaming `tenantId` to `tenant` I'll discuss it with BE devs and rename this arg in hooks

mariia-aloshyna: Created a PR in stripes-core - https://github.com/folio-org/stripes-core/pull/1348

zburke: I brought up `tenant` vs `tenantId` with BE folks already and was basically told, "This is already a mess in other places. Changing it here won't actually make the whole system less messy, it'll just make it less messy for you. Since our work is already done and `tenantId` is a better name anyway, we're not gonna do it again. Sorry not sorry." 😒

mariia-aloshyna: yep, I noticed that mod-search also uses `tenantId` prop e.g.

PR #1480, opened 2023-10-17
UIDATIMP-1544 "Invalid Date" appearing on completed file upload cards.

JohnC-80:
## Problem
With the file-splitting feature active, an uploadedDate value is created when an upload is complete. This date was being locale-formatted before being handed to an instance of `react-intl`'s `<FormattedDate>` - sometimes it will parse the string correctly, sometimes it will not depending on the separator that is used or the order of day, month, year. Some browser locale settings may see the words "Invalid Date" rather than an actual date value.

## Approach
use Date().toISOString() instead of Date().toLocaleDateString() in multipartUploader so that `<FormattedDate>` receives a consistently formatted string instead of a string that can vary based on browser locale.


ncovercash (approved):
Good catch!

PR #1483, opened 2023-10-19
STRIPES-854 ⚡⚡⚡ use all available CPUs in CI

zburke:
* Jest does not automatically detect the number of CPU cores available in CI and hence runs single threaded. Manually checking for multiple CPUs allows it to run multi-threaded, reducing test time from > 13m to ~9m. 
* Align `stripes` deps: you must choose between `~` and `^`; mixing them leads to inconsistencies and dupes. `~` locks to the Poppy versions, which is not desirable long-term since it will put this application out of sync with others in folio-snapshot that have already moved onto Quesnelia, but was the shortest path to passing tests right now.
* Suppress RTL error output related to misconfigured tests to make test output more manageable. We must file a separate Jira before accepting this.

Refs [STRIPES-854](https://issues.folio.org/browse/STRIPES-854)


mariia-aloshyna (approved):

mariia-aloshyna:
Thank you @zburke for figuring that out!

JohnC-80 (approved):

zburke:
@mariia-aloshyna, @JohnC-80 : I cannot get the axe tests to run reliably here. I thought maybe the problem was running in parallel, but I see failures in CI whether it runs them in serial or parallel. I also don't understand why the tests take longer in CI than when I run them locally. 

The only thing I can think of is to skip all the axe runs, but that feels awful. I am swamped under PR reviews and RTR work and don't really have time to investigate this further. If you don't want to pursue the CI changes, somebody at least needs to rescue the "use consistent stripes version constraints" work and apply it in another PR; that's the most important thing happening here because you'll get duplicate stripes-* things in your build without that fix.

mariia-aloshyna:
Hi @zburke I created [PR](https://github.com/folio-org/ui-data-import/pull/1484) to align versions, and it looks like the tests there passed successfully

zburke:
Closing in favor of #1484, which has the magic sauce to get tests to pass. 👏 👏 👏 

It seems that parallel execution and axe tests just don't get along well here.

PR #1494, opened 2023-11-02
UIDATIMP-1563 - remove failed jobs listing form composite job details. Include failed jobs in part calculation.

JohnC-80:
https://issues.folio.org/browse/UIDATIMP-1563

It was difficult to differentiate between "failed" jobs and jobs "completed with errors" from the level of Composite Jobs, so the overlap in numbers was confusing for users. We opted to just remove the failed jobs line from the card details altogether.

Also, the 'Job parts processed' field was not accounting for failed jobs in its 'processed' number, so now those number make better sense since both processed and 'error' status jobs are considered.

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs fixed:
 * The '[number of records] - records found' subtitle is displayed after clicking on the textLink error counter (UIDATIMP-1549)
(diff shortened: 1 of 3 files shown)


ncovercash (approved):
LGTM! Thanks for your quick work on these

mariia-aloshyna (approved):

PR #1500, opened 2023-11-08
UIDATIMP-1566: Add maxUnpagedResourceCount limit from stripes & increase PER_REQUEST_LIMIT to 3000

OleksandrHladchenko1:
## Purpose

* The locations query contains a limit=1000 clause, causing some locations to be unassignable in tenants with more than 1000 entries.

## Approach

* Created `getWrapperSourceLink` function that accept `key` and tenant's limit and returns a path with tenant's limit or with `PER_REQUEST_LIMIT` value
* Increase `PER_REQUEST_LIMIT` variable from 1000 to 3000.
* Get `requestLimit` prop from `stripes.config.maxUnpagedResourceCount` in `MappingProfilesForm` component and pass it to the components which use `AcceptedValuesField` with `wrapperSourceLink` prop to get the data.
* Fixed tests in `getIdentifierTypes` method.

## Refs

[UIDATIMP-1566](https://issues.folio.org/browse/UIDATIMP-1566)


OleksandrHladchenko1:
@folio-org/fe-tl-reviewers please take a look.

mariia-aloshyna (approved):

BogdanDenis (approved):

PR #1541, opened 2024-01-30
UIDATIMP-1592: Fix a crash when viewing Authority DELETE job mapping profile view.

Dmytro-Melnyshyn:
## Purpose
Fix a crash when viewing Authority DELETE job mapping profile view.


zburke (approved):
LGTM

NB: commit subjects/PR descriptions should describe the solution (e.g. "handle empty marc mapping label") rather than the problem ("fix problem..."). With the latter, we end up with a commit log that just looks like 
* fix bug ...
* fix bug ...
* fix bug ...

Comments should describe the code, but the bug is no longer in the code once the PR is merged!

Dmytro-Melnyshyn:
> LGTM
> 
> NB: commit subjects/PR descriptions should describe the solution (e.g. "handle empty marc mapping label") rather than the problem ("fix problem..."). With the latter, we end up with a commit log that just looks like
> 
> * fix bug ...
> * fix bug ...
> * fix bug ...
> 
> Comments should describe the code, but the bug is no longer in the code once the PR is merged!

@zburke you are right, the solution is much better. I've changed.

BogdanDenis (approved):

PR #1554, opened 2024-03-13
UIDATIMP-1600: Use `incomingRecordId` to retrieve Incoming record logs

OleksandrHladchenko1:
## Purpose

* Currently, we use `sourceRecordId` to get logs for the incoming record tab, it is needed to use `incomingRecordId` instead.

## Approach

* Pass `inconingRecordId` as the url's parameter.
* 

## Refs

[UIDATIMP-1600](https://folio-org.atlassian.net/browse/UIDATIMP-1600)


OleksandrHladchenko1:
@folio-org/fe-tl-reviewers please review

mariia-aloshyna:
I don't think it's a good idea to put incomingRecordId to URL params, I would prefer to retrieve it from the response, but I see that the `metadata-provider/jobLogEntries/${logId}/records/${recordId}` endpoint doesn't return the `incomingRecordId` field. Let me discuss it with BE

Dmytro-Melnyshyn (approved):

mariia-aloshyna (approved):

PR #1555, opened 2024-03-22
Release v7.1.0

mariia-aloshyna:


zburke (commented):
Dropping support for an Okapi interface version, like dropping support for a peer dependency version, even if it is a minor, is a breaking change because you are saying, "This application has new requirements; it no longer functions in the environment where it used to run just fine."

PR #1560, opened 2024-04-05
UIDATIMP-1603: "addedRelations" is not clearing after unlinking the update profile

mariia-aloshyna:
## Purpose

If the user creates a job profile and links the action profile with the update action and after this unlinks it and tries to save the job - the error “ActionProfile with id [ID] and action UPDATE requires linked MatchProfile" will appear.

## Approach

- Check if "reactTo" is null or undefined before comparison.

## Refs

https://folio-org.atlassian.net/browse/UIDATIMP-1603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change history for ui-data-import
 
+## **7.1.1** (in progress)
+
+### Bugs fixed:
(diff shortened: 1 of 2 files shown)


OleksandrHladchenko1 (commented):
Fix eslint warnings please

OleksandrHladchenko1 (approved):

mariia-aloshyna:

vashjs (approved):

PR #1561, opened 2024-04-09
UIDATIMP-1605: Allow to search profiles using `*`

OleksandrHladchenko1:
## Purpose

* When user try to search profile using `*` (or just enter `*` and try to search) he gets the error
* There are templates in profiles' manifests (e.g. `name="%{query.query}*"`). It already has the necessary asterisk for BE. So we need to remove the other asterisks in order to avoid the error message.

## Approach

* Created function `validateSearchString` that removes `*`
* Added check in `SearchAndSort` component (when user search the profile)
* Added check in `getSearchQuery` function (when user search the profile while linking)

## Refs

[UIDATIMP-1605](https://folio-org.atlassian.net/browse/UIDATIMP-1605)

### Before
https://github.com/folio-org/ui-data-import/assets/85172747/a46f3eaf-5551-43a8-bcf1-3ba66b7d60c5

### After
https://github.com/folio-org/ui-data-import/assets/85172747/6513f49c-fb7d-4bf0-ab33-95701a4a257f


mariia-aloshyna (approved):

OleksandrHladchenko1:
@folio-org/fe-tl-reviewers could you please review this PR?

artem-blazhko (approved):

PR #1630, opened 2024-10-10
UIDATIMP-1666: Display correctly `Percentage` and `Currency` fields for orders and invoices

OleksandrHladchenko1:
## Purpose

* Sometimes we get percentage and currency types in format **percentage** or **Percentage**.

## Approach

* Added `.toLowerCase()` for amountType.
* Added unique id for **Donor information** accordion.

## Refs

[UIDATIMP-1666](https://folio-org.atlassian.net/browse/UIDATIMP-1666)


mariia-aloshyna (approved):

src/settings/MappingProfiles/detailsSections/utils.js
```diff
@@ -241,10 +241,10 @@ export const getAccountNumbersByCode = organization => {
   }), {});
 };
 
-export const PERCENTAGE_VALUE = 'Percentage';
+export const PERCENTAGE_VALUE = 'percentage';
 
 export const renderAmountValue = (amountValue, amountType, currency) => {
-  const amountSymbol = amountType === `"${PERCENTAGE_VALUE}"` ?
+  const amountSymbol = amountType?.toLowerCase() === `"${PERCENTAGE_VALUE}"` ?
```
zburke, line 247: `PERCENTAGE_VALUE` is already a string so string interpolation is unnecessary:
```
const amountSymbol = amountType?.toLowerCase() === PERCENTAGE_VALUE ?
```

Also, you don't need to change this now since you have working code, but it does a lot of reinventing the wheel: everything you're doing here and in `<CurrencySymbol>` is already handled by [react-intl's FormattedNumber](https://formatjs.io/docs/react-intl/components#formattednumber). Just use it.

zburke (approved):

src/settings/MappingProfiles/detailsSections/utils.js
mariia-aloshyna: @zburke that's data import mapping syntax specific, everything that is not a MARC field like '918$a' should be wrapped in the quotes

mariia-aloshyna: on UI it looks like it for example, otherwise, you will get an error, because it will check for a mapping syntax
Screenshot 2024-10-11 at 19 55 51

zburke: 🤯

PR #1631, opened 2024-10-14
UIDATIMP-1669: Bump stripes to v9.2.0 for Ramsons

OleksandrHladchenko1:
## Purpose
* Bump `stripes` to `v9.2.0` and [all stripes related modules](https://github.com/folio-org/stripes/releases/tag/v9.2.0) for Ramsons release.

## Links
[UIDATIMP-1669](https://folio-org.atlassian.net/browse/UIDATIMP-1669)


mariia-aloshyna (approved):

SergeYvas (approved):

OleksandrHladchenko1:
> Technically bumping a peer is always a breaking change since you are stating a new external requirement. If your tests will pass with stripes-* dev-deps pinned to their stripes v9.1 versions, i.e. this app does not require any of the new features added to stripes in v9.2, you can leave the peer-dep and this repo's major version as-is and this PR is unnecessary.
> 
> Alternatively, if you choose to keep the new 9.2.0 requirement in peer-deps, then you need to bump this repo's version from 7.1.8 to 8.0.0.

Hi @zburke  We decided to keep the new `9.2.0` version of `stripes` in `peerDependencies`. Therefore we bumped the major version of `ui-data-import`.

PR #1655, opened 2024-11-20
UIDATIMP-1634: Refactor away from default props for functional components

OleksandrHladchenko1:
## Purpose
* `defaultProps` will be removed from function components in place of ES6 default parameters. Class components will continue to support `defaultProps` since there is no ES6 alternative.

## Approach
* Removed `defaultProps` in `ViewJobProfileComponent` component

## Refs
[UIDATIMP-1634](https://folio-org.atlassian.net/browse/UIDATIMP-1634)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change history for ui-data-import
 
-## [8.0.2] (IN PROGRESS)
+## [8.1.0] (IN PROGRESS)
+
+### Features added:
+* `React v19`: refactor away from default props for functional components. (UIDATIMP-1634)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Change history for ui-data-import
 
-## [8.0.2] (IN PROGRESS)
+## [8.1.0] (IN PROGRESS)
```
mariia-aloshyna, line 3: (UIDATIMP-1675) should be still under 8.0.2, it's gonna be released as ramsons bf

mariia-aloshyna (approved):

PR #1661, opened 2024-12-04
UIDATIMP-1686: Replace `_/proxy/tenants/${tenant}/modules` request with `stripes.discovery.modules` object

mariia-aloshyna:
## Purpose

While troubleshooting an issue in the Eureka-bugfest environment, we ran across a [bit of code in ui-data-import](https://github.com/folio-org/ui-data-import/blame/f7b36a39ac5db4e3a5d71f42f120f101e97f1009/src/settings/MatchProfiles/MatchProfiles.js#L240-L249) that queries Okapi for the tenant’s currently-installed modules with a manifest like:
`_/proxy/tenants/${tenant}/modules`;
This same data is available on stripes at stripes.discovery.modules . The discovery process in Eureka is quite different than in legacy FOLIO, with different endpoints etc etc, but all that logic is consolidated within stripes-core at present. If we can leverage the discovery data attached to stripes instead of making a query, then we can keep it that way and avoid the need for Eureka-specific code to pollute ui-data-import.

## Approach

- Get rid of `_/proxy/tenants/${tenant}/modules` request in Match profiles and get data from `stripes.discovery.modules` object.

## Refs

https://folio-org.atlassian.net/browse/UIDATIMP-1686

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Bugs fixed:
 * Allow central tenant to create action profile for Orders and Invoices. (UIDATIMP-1679)
 * Allow central tenant to create filed mapping profile for Orders and Invoices. (UIDATIMP-1685)
+* Replace `_/proxy/tenants/${tenant}/modules` with `stripes.discovery.modules` object. (UIDATIMP-1686)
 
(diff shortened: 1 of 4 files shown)


zburke (approved):
🎉 ❤️

OleksandrHladchenko1 (approved):

PR #1663, opened 2024-12-06
UIDATIMP-1689: Populate `Existing records` field when create match profile

OleksandrHladchenko1:
## Purpose

* List of all existing record fields should be displayed.

## Approach

* `!isEqual(prevProps.stripes.discovery.modules, this.props.stripes.discovery.modules` is always `false`. That's why we never fetch JSON schemas.
* Fetch options only once and check if `this.props.stripes.discovery.modules` is not empty in order to avoid failed requests. 

## Refs

[UIDATIMP-1689](https://folio-org.atlassian.net/browse/UIDATIMP-1689)

## Screenshots

image
image


mariia-aloshyna (approved):

UladzislauKutarkin (approved):

zburke (commented):
I need more information about why moving from cDU to cDM fixes this bug. It should not be on the reviewer to triage the bug and assess how the fix works; you need to explain that in the PR description.

PR #1677, opened 2025-02-13
UIDATIMP-1702: Replace dependency on mod-configuration with mod-settings

OleksandrHladchenko1:
## Links
[UIDATIMP-1702](https://folio-org.atlassian.net/browse/UIDATIMP-1702)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features added:
 * `React v19`: refactor away from default props for functional components. (UIDATIMP-1634)
+* *BREAKING* Replace dependency on `mod-configuration` with `mod-settings`. (UIDATIMP-1702)
 
 ## [8.0.5](https://github.com/folio-org/ui-data-import/tree/v8.0.5) (2024-12-19)
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/data-import",
-  "version": "8.0.5",
+  "version": "8.1.5",
   "description": "Data Import manager",
   "main": "src/index.js",
   "repository": "folio-org/ui-data-import",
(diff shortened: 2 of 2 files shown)


zburke (changes requested):
Thank you for calling out this breaking change in the CHANGELOG. Please make the corresponding change in `package.json`. 

Reading the [mod-settings documentation](https://github.com/folio-org/mod-settings/blob/a0c26931bc705b30197ee17e97549e638f273b26/doc/HOWTO.md#summary), I think the permissions you are granting are necessary but not sufficient. It states, 
> users must be assigned whatever permissions defined by the client module are needed for the settings access they are intended to have ... In addition to these, every user must be assigned basic permissions to use the settings modules. 

Here, you have only granted the "basic permissions" that allow searching:
```
mod-settings.entries.collection.get
```
but not the specific permissions that allow values in a given scope to be retrieved. IIUC, this means you also need to grant permission in the matching scopes:
```
mod-settings.global.read.tenant
mod-settings.global.read.orders
```
But, I am also not confident these scopes are correct. For example, [in ui-tenant-settings I see the following](https://github.com/folio-org/ui-tenant-settings/blob/dadc103cdbc62ad8323b1c4e36aae35b0a3d6683/package.json#L41):
```
mod-settings.global.read.ui-tenant-settings.addresses.manage
```
There is nothing analogous in ui-orders. Has that work been defined? Should it/must it be completed first?

src/settings/MappingProfiles/detailsSections/edit/OrderDetailSection/OrderInformation.js
```diff
@@ -482,20 +483,20 @@ const OrderInformationComponent = ({
 OrderInformationComponent.manifest = Object.freeze({
   purchaseOrderLinesLimitSetting: {
     type: 'okapi',
-    path: 'configurations/entries?query=(module==ORDERS and configName==poLines-limit)',
+    path: 'orders-storage/settings?query=(key==poLines-limit)',
```
zburke: This implies a dependency on `orders-storage.settings`, but this is absent from `package.json::stripes.okapiInterfaces`.

zburke (changes requested):

PR #1681, opened 2025-02-27
UIDATIMP-1712 apply wrappers with flow()

zburke:
Replace annotations such as `@withStripes` with a [`_.flow()`](https://lodash.com/docs/4.17.15#flow)-based syntax for better compatibility with esbuild-loader >= 3.1.0.

Refs [UIDATIMP-1712](https://folio-org.atlassian.net/browse/UIDATIMP-1712)


src/components/DataFetcher/DataFetcher.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import {
+  flow,
```
mariia-aloshyna, line 4: hi @zburke there is an unused import

JohnC-80 (approved):

PR #1689, opened 2025-03-17
UIDATIMP-1716: Place wrappers in right order to not lose any data after applying `flow()`

mariia-aloshyna:
## Purpose

The 'Something went wrong' error message is displayed after clicking on existing job/match/action/mapping profiles.

## Approach

- Place wrappers in right order to not lose any data after migrating to `flow()`.

## Refs

https://folio-org.atlassian.net/browse/UIDATIMP-1716

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change history for ui-data-import
 
+## [9.0.1] (IN PROGRESS)
+
+### Bugs fixed:
+* Place wrappers in right order to not lose any data after applying `flow()`. (UIDATIMP-1716)
+
 ## [9.0.0](https://github.com/folio-org/ui-data-import/tree/v9.0.0) (2025-03-13)
 
 ### Features added:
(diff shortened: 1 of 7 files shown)


BogdanDenis (approved):

OleksandrHladchenko1 (approved):

mariia-aloshyna:
> 😱 this resolves bugs introduced in #1681 and #1685. Mea culpa, mea culpa. I am so sorry for introducing these bugs 😢 . 
> 
> 
> 
> But also ... why did behavior change with `_.flow()`? Those PRs very specifically preserved the order of the wrappers. 🤨 I'm glad you identified the problem, but I would like to understand the solution better. Any context you can provide would be helpful.

Hi @zburke no worries!
Decorators are applied from bottom to top, so the last HOC applied wraps all the previous ones. But 'flow()' works from left to right, that means that we need to take the very bottom decorator and use it as a first function in 'flow()' if we want to use it instead of decorators, this is the sequence

PR #1696, opened 2025-05-27
UIDATIMP-1725: Fix stuck Progress Bar in import files

OleksandrHladchenko1:
## Purpose

* Unblock progress of importing file, that was created via API.

## Approach

* `currentProgress` should be compared with `total`, not 100. It works wrong when user imports more than 100 records.

## Refs

[UIDATIMP-1725](https://folio-org.atlassian.net/browse/UIDATIMP-1725)


OleksandrHladchenko1:
@zburke @mariia-aloshyna 
I think this check will help us with wrong data from BE
`const currentProgress = current <= total ? current : total;`

If `current > total` then `current` will be `total` and it will show `100%` in the progress bar. Otherwise we will see the default behavior for the progress bar.
`current` shows us the number of already uploaded records, not the percentage. So, setting `currentProgress = 100` may be wrong if we try to upload more than 100 records (`total`  is 100+).
Correct me please if I'm wrong

PR #1748, opened 2026-05-29
UIDATIMP-1714: Replace moment with day.js

mariia-aloshyna:
## Purpose

Applications should import dayjs from stripes-components and do not need to maintain their own dependency.

## Approach

- Replace `moment` with `dayjs` from stripes-components

## Refs

https://folio-org.atlassian.net/browse/UIDATIMP-1714

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [10.1.0] (IN PROGRESS)
 
+### Features added:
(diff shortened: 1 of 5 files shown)


OleksandrHladchenko1 (approved):

BogdanDenis (approved):

zburke (approved):
🙌

PR #1749, opened 2026-06-30
UIDATIMP-1764 always navigate to unguardable routes

zburke:
Sometimes, navigation cannot be guarded (as when navigating to `/logout` when a session times out). Call `stripes-core::isGuardable()` to determine whether the route may be guarded or if navigation is obligatory.

Do not show a "Stay here?" prompt for those obligatory routes.

Refs [UIDATIMP-1764](https://folio-org.atlassian.net/browse/UIDATIMP-1764), [STRIPES-1026](https://folio-org.atlassian.net/browse/STRIPES-1026)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Change history for ui-data-import
 
-## [10.1.0] (IN PROGRESS)
+## [11.0.0] (IN PROGRESS)
 
 ### Features added:
 * Replace moment with day.js (UIDATIMP-1714)
 
 ### Bugs fixed:
 * Fix translation for select placeholder. (UIDATIMP-1756)
+* Always allow redirect to `/logout` on session termination. (UIDATIMP-1764)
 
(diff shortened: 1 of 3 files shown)


JohnC-80 (approved):

zburke:
@OleksandrHladchenko1, please take a look at this PR in light of the related stories [STCOR-1076](https://folio-org.atlassian.net/browse/STCOR-1076) and [STRIPES-1026](https://folio-org.atlassian.net/browse/STRIPES-1026). We started to pursue this by modifying how XHR works, but eventually concluded that when a session times out in the middle of an XHR request, the correct behavior is for logout to proceed, and therefore that the change should be in DI rather than in stripes-core. Note that [stripes-form](https://github.com/folio-org/stripes-form/pull/188) and [stripes-final-form](https://github.com/folio-org/stripes-final-form/pull/110) already had hard-coded exceptions for the `/logout` route, and that as part of this STRIPES ticket they, too, have been modified to leverage the `isGuardable()` export from stripes-core. 

Please let us know if you have any questions.

OleksandrHladchenko1 (approved):

PR #1753, opened 2026-07-21
UIDATIMP-1731: Create a capability: Data import: Delete authority records

OleksandrHladchenko1:
The result will be:
* `UI-Data-Import Authority-Records-Delete — Execute (Procedural)`

## Link
[UIDATIMP-1731](https://folio-org.atlassian.net/browse/UIDATIMP-1731)

### DIFF
--- a/package.json
+++ b/package.json
@@ -451,6 +451,12 @@
           "tags.collection.get"
         ],
         "visible": true
+      },
+      {
+        "permissionName": "ui-data-import.authority-records-delete.execute",
+        "displayName": "Data import: Delete authority records",
+        "visible": true,
+        "subPermissions": []
       }
     ]
   }
--- a/package.json
+++ b/package.json
@@ -453,7 +453,7 @@
         "visible": true
       },
       {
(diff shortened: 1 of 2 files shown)


BogdanDenis (approved):

package.json
```diff
@@ -451,6 +451,12 @@
           "tags.collection.get"
         ],
         "visible": true
+      },
+      {
+        "permissionName": "ui-data-import.authority-records-delete.execute",
```
zburke: This isn't wrong, but it is sort of odd. Could we use `...authority-records.delete`? It's compliant with [naming conventions](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/156368925/Permissions+naming+convention#5.-Action-Verb), shorter, and more focused. I understood `.execute` to be used for things like `check-in-by-barcode.execute` where `.create` feels inappropriate.

zburke (approved):
Should there be a corresponding code change to actually implement the given scenarios?

